PR #17, opened 2020-06-26
Create LIP2: Libra roles and permissions

sblackshear:


lips/lip-2.md
```diff
@@ -0,0 +1,1430 @@
...
+This LIP describes the conceptual model and implementation of access control on the Libra blockchain.
+
+---
+# Abstract / Motivation
+---
+
+Libra uses a variant of role-based access control (RBAC) to restrict access to sensitive on-chain operations.
+
+A *role* is an entity with some authority in the Libra system. The roles in the initial system are:
+LibraRoot - the root authority of Libra. Controlled jointly by Libra Payment Networks and the Association Council
```
dahliamalkhi: Please add formatting (specifically add bullets here), line breaks disappear in md

lips/lip-2.md
```diff
@@ -0,0 +1,1430 @@
...
+ParentVASP - The primary account of a regulated wallet operating on the Libra blockchain
+ChildVASP - A subaccount of a particular parent VASP operating on the Libra blockchain
+
+Every account in the Libra system is created with a single, immutable role that is granted at the time the account is created. Creating an account with a particular role is a privileged oepration (e.g., only an account with the ParentVASP role can create an account the ChildVASP role). In some cases, the role is globally unique (e.g., there is only one account with the LibraRoot role). In other cases, there may be many accounts with the given role (e.g., ChildVASP).
+
+A *permission* is the authority to perform a sensitive action in the Libra system. Each permission may be assigned to one or more roles (usually only one), and each role may have zero or more permissinons. Permissions can be assigned in genesis, upon account creation (most common), or claimed by an existing account with the appropriate role. Like roles, some permissions are globally unique (e.g., the permission to mint a particular currency type) and some are not.
```
dahliamalkhi, line 32: if there are multiple account for some role (eg ParentVASP), is a permission granted to the role (ParentVASP) or to an account instance of that role (eg coinbase)?

JoelMarcey:
Hi @sblackshear - yeah, it looks like Docusaurus is not happy with the formatting of your markdown file. See https://app.netlify.com/sites/libra-lip/deploys/5f10e29beaaf540008748ae0

I think for some, maybe even the majority, of the places you use raw HTML, you could use markdown syntax. Did you find that it wasn't working?

lips/lip-2.md
sblackshear: The latter--e.g., there is one account with a `ParentVASP` role per real-world VASP.

sblackshear:
> Hi @sblackshear - yeah, it looks like Docusaurus is not happy with the formatting of your markdown file. See https://app.netlify.com/sites/libra-lip/deploys/5f10e29beaaf540008748ae0
> 
> I think for some, maybe even the majority, of the places you use raw HTML, you could use markdown syntax. Did you find that it wasn't working?

Unfortunately, the raw HTML comes from the autogenerated [doc](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Roles.md) for the Roles module. I was hoping I could use that directly, but it looks like I either need to manually edit it or fix the doc generator first...

JoelMarcey:
Ah - didn't realize they were generated. Maybe let's dig into where the markdown parser is struggling and see if we can modifier your generator to output something more consumable. 

Barring that, maybe we can see if we can update the markdown parser, but that would be a bit more work maybe.

Maybe first, though, we try to fix manually to see how many times the markdown parser struggles - if it is only a few, then we can at least edit manually and get your PR merged while we see if we can make any long term fixes.

sblackshear:
Created this one in GH UI and couldn't figure out how to update in my local repo--see https://github.com/libra/lip/pull/24 instead.

PR #18, opened 2020-06-27
Transaction metadata specification

kphfb:
Describe standardized format for transaction metadata

### DIFF
--- a/lips/lip-4.md
+++ b/lips/lip-4.md
@@ -1,0 +1,222 @@
+---
+lip: 4
+title: Transaction Metadata Specification
+author: Kevin Hurley (@kphfb)
+status: Draft
+type: Informational
+created: 06/26/2020
+---
+# Summary
+---
+Merchants and wallet holders may rely on third parties to maintain custody of their wallet and may optionally want to receive additional data in transactions.  Libra leverages subaddressing to provide this functionality.
+
+---
+# Terminology
+___
+*subaddress:* Accounts on-chain are represented by an address.  To allow multiplexing of a single address into multiple wallets, custodial wallets may utilize “subaddresses” under the on-chain address for each underlying user.  These subaddresses represent the users and may have meaning only to the custodian - but the mapping from subaddress to user is not known on-chain, but rather is an internal mapping known by the custodian.
+*referenced_event:* In the case where funds must be returned, referenced_event refers to the event sequence number of the sender’s original sent payment event.  Since refunds are just another form of p2p transfer, the referenced event field allows a refunded payment to refer back to the original payment.
+
+
+---
+# Abstract / Motivation
+---
+
+Merchants and wallet holders may rely on third parties to maintain custody of their wallet and may optionally want to receive additional data in transactions.  Libra leverages subaddressing to provide this functionality.  This document describes an approach to supporting transactions like this by defining a standard that supports:
+
+* Subaddressing to support multiplexing a single account address into multiple wallets, enabling custodial wallets to maintain a single address rather than maintain numerous addresses.
+
+As an example, assume a recipient account provides custodial wallets. To send funds to this wallet within Libra:
+
+* The recipient crafts a QR code containing a URI that references the account hosting the custodial wallet, or the host, and a reference to their wallet.
+* The consumer scans the QR code and constructs a transaction and includes within it the subaddress.
+* Once the transaction has been submitted and committed to the blockchain, the host notifies the recipient of the transaction.
+
+---
+# Specification
+---
+
+# The Lifetime of a Transaction Containing Metadata
+
+The first step to submitting a transaction is producing the metadata. The sender first produces a *Libra Canonically Serialized (LCS)* metadata_wrapper consisting of an LCS-serialized MetadataType object:
+
+```
+enum MetadataType {
+  Undefined,
+  GeneralMetadataType(GeneralMetadata),
+  TravelRuleMetadataType(TravelRuleMetadata),
+  UnstructuredStringMetadataType(UnstructuredStringMetadata),
+}
+
+// Used for versioning of general metadata
+enum GeneralMetadata {
+    GeneralMetadataVersion0(GeneralMetadatav0),
+}
+
+struct GeneralMetadatav0 {
+    // Subaddress to which the funds are being sent
+    Option<Vec<u8>> to_subaddress,
+    
+    // Subaddress from which the funds are being sent
+    Option<Vec<u8>> from_subaddress,
+    
+    // Event sequence number of referenced payment
+    Option<u64> referenced_event,
+}
+
+// Used for versioning of travel rule metadata
+enum TravelRuleMetadata {
+    TravelRuleMetadataVersion0(TravelRuleMetadatav0),
+}
+
+struct TravelRuleMetadatav0 {
+    // Off-chain reference_id.  Used when off-chain APIs are used.
+    // Specifies the off-chain reference ID that was agreed upon in off-chain APIs.
+    Option<String> off_chain_reference_id,
+}
+
+struct UnstructuredStringMetadata {
+    // Unstructured string metadata
+    Option<String> metadata,
+}
+```
+
+
+Using the initial example described in the motivation, the merchant whose wallet, 0x5678, is hosted by 0x1234 may post a URI for shoes that cost 20 microlibra.  The purchaser would then submit a transaction containing the following metadata:
+
+```
+0x01, 0x00, 01, 0x5678, 00, 00, 00
+/* general_metadata_type, general_metadata_v0, 
+      to_subaddress_present, to_subaddress,  
+      from_subaddress_not_present, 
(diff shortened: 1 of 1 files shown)


lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+created: 06/26/2020
+---
+# Summary
+---
+Merchants and wallet holders may rely on third parties to maintain custody of their wallet and may optionally want to receive additional data in transactions.  Libra leverages subaddressing to provide this functionality.
+
+---
+# Terminology
+___
+*subaddress:* Accounts on-chain are represented by an address.  To allow multiplexing of a single address into multiple wallets, custodial wallets may utilize “subaddresses” under the on-chain address for each underlying user.  These subaddresses represent the users and may have meaning only to the custodian - but the mapping from subaddress to user is not known on-chain, but rather is an internal mapping known by the custodian.
```
sblackshear, line 16: It would be good to introduce subaddresses in a way that discourages reuse. For example, we can say that the best practice is a many-to-one subaddress -> user mapping (where ideally "many" means "per payment"). If we really want to discourage reuse, we could describe a subaddress as an identifier for a specific payment rather than an identifier for a user (though of course some use-cases like refunds will violate this).

The current description makes it sounds like there might be a 1-1 mapping between subaddresses and users might. This is a choice VASPs can make, but one that we should (I think) advise against.

sblackshear: Formatting issue here (`referenced_event` on same line).

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+*referenced_event:* In the case where funds must be returned, referenced_event refers to the event sequence number of the sender’s original sent payment event.  Since refunds are just another form of p2p transfer, the referenced event field allows a refunded payment to refer back to the original payment.
```
sblackshear, line 17: sender's original -> original sender's might be slightly clearer

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+
+Merchants and wallet holders may rely on third parties to maintain custody of their wallet and may optionally want to receive additional data in transactions.  Libra leverages subaddressing to provide this functionality.  This document describes an approach to supporting transactions like this by defining a standard that supports:
+
+* Subaddressing to support multiplexing a single account address into multiple wallets, enabling custodial wallets to maintain a single address rather than maintain numerous addresses.
+
+As an example, assume a recipient account provides custodial wallets. To send funds to this wallet within Libra:
+
+* The recipient crafts a QR code containing a URI that references the account hosting the custodial wallet, or the host, and a reference to their wallet.
+* The consumer scans the QR code and constructs a transaction and includes within it the subaddress.
+* Once the transaction has been submitted and committed to the blockchain, the host notifies the recipient of the transaction.
```
sblackshear, line 32: "and credits the off-chain account associated with the subaddress" (or similar)?

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+# Abstract / Motivation
+---
+
+Merchants and wallet holders may rely on third parties to maintain custody of their wallet and may optionally want to receive additional data in transactions.  Libra leverages subaddressing to provide this functionality.  This document describes an approach to supporting transactions like this by defining a standard that supports:
+
+* Subaddressing to support multiplexing a single account address into multiple wallets, enabling custodial wallets to maintain a single address rather than maintain numerous addresses.
+
+As an example, assume a recipient account provides custodial wallets. To send funds to this wallet within Libra:
+
+* The recipient crafts a QR code containing a URI that references the account hosting the custodial wallet, or the host, and a reference to their wallet.
```
sblackshear, line 30: Might be clearer to avoid introducing the term "host" and just keep saying "custodial wallet" or "custodian"

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+}
+
+struct GeneralMetadatav0 {
+    // Subaddress to which the funds are being sent
+    Option<Vec<u8>> to_subaddress,
+    
+    // Subaddress from which the funds are being sent
+    Option<Vec<u8>> from_subaddress,
+    
+    // Event sequence number of referenced payment
```
sblackshear, line 62:
```suggestion
    // Event sequence number of the `SentPaymentEvent` that this transaction is refunding
```

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+}
+
+struct UnstructuredStringMetadata {
+    // Unstructured string metadata
+    Option<String> metadata,
+}
+```
+
+
+Using the initial example described in the motivation, the merchant whose wallet, 0x5678, is hosted by 0x1234 may post a URI for shoes that cost 20 microlibra.  The purchaser would then submit a transaction containing the following metadata:
```
sblackshear, line 84: - I didn't understand what 0x1234 was and why it didn't appear in the metadata
- 0x5678 got a screamin' deal on those shoes

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+
+```
+0x01, 0x00, 01, 0x5678, 00, 00, 00
```
sblackshear, line 87: This is a bit hard to read. I'm not sure if it's much better but something like

```
0x01 // general_metadata_type
0x00 // general_metadata_v0
...
```
might be easier.

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+/* general_metadata_type, general_metadata_v0, 
+      to_subaddress_present, to_subaddress,  
+      from_subaddress_not_present, 
+      referenced_event_not_present */
+```
+
+## Submitting a transaction
+
+With the metadata in hand, the sender can now submit a transaction to the Libra block chain via a deposit call in Move:
```
sblackshear, line 96: We should probably link to the P2P [script](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/doc/peer_to_peer_with_metadata.md) here instead, since that is what clients will actually put in a tx.

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+
+```
+deposit<LBR>(
+    recipient_address, 
+    amount, 
+    metadata, 
+    metadata_signature);
+```
+
+In the case of the example above, 0x1234 would be the recipient and 20 LBR would be the amount.  Note that metadata_signature must only be present for travel-rule cases between VASPs.
```
sblackshear, line 106: - Ah, now I see what 0x1234 is now. We might want to use different syntactic conventions for Libra addresses and subaddresses to avoid confusion (e.g., don't use 0x for the subaddress).
- Might want to say: `metadata` and `metadata_signature` are optional for transactions not subject to the travel rule (unless it sets a bad precedent to say the `metadata` is optional)

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Processing the transaction
+
+Much like any other funds transfer request, validators only verify that the sender has the sufficient libra to support the transaction and that the transaction is properly formed and valid, they do not inspect or verify the correctness of the metadata.
```
sblackshear, line 110: Should we say what a custodian should do with malformed metadata (presumably refund?)

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+RawTransaction {
+    sender_account: 0x7777,
+    program: program, 
+}
+```
+
+
+## Refunds
+
+Merchant B now wishes to refund user A. But user A was sending from a custodial account so merchant B must send the funds back to the custodial account and include subaddress information so that the funds are directed back to user A.  Merchant B’s client now constructs a transaction via the following where referenced_event is the committed event sequence number under the sending account of the original send payment event:
```
sblackshear, line 167:
```suggestion
Merchant B now wishes to refund user A. But user A was sending from a custodial account so merchant B must send the funds back to the custodial account and include subaddress information so that the funds are directed back to user A.  Merchant B’s client now constructs a transaction via the following where referenced_event is the committed event sequence number under the sending account of the original sent payment event:
```

lips/lip-4.md
```diff
@@ -0,0 +1,222 @@
...
+lcs_metadata = lcs(MetadataType::TravelRuleMetadataType(
+        TravelRuleMetadata::TravelRuleMetadataVersion0(metadata))),
+
+// The receiver side will have signed this as part of the off-chain APIs
+// and will have sent it to the sender side
+{ // Done by receiver VASP
+    metadata_signature = sign(metadata, sender_address, amount, receiver_key);
+}
+
+program = deposit<LBR>(
```
sblackshear, line 210:
```suggestion
encode_peer_to_peer_with_metadata_script("LBR" /*currency type*/,
```

lips/lip-4.md
kphfb: Yeah, the signature is only for TR, but the metadata object itself isn't necessarily related to travel rule always.  You could put a note in there if you wanted.  You could put a refund ID in there.  Let me try to make it clearer

PR #19, opened 2020-07-02
Update for travel rule soft-match/manual review

kphfb:
Be able to specify that we are in manual review state. Then allow manual review state to specify that we have a soft match and other side can provide free-form KYC data.

This also results in changing the KYC data object status field to no longer be a pure enum but instead have the ability to describe why we aborted. This is also probably useful for other cases where we abort


lips/lip-1.md
```diff
@@ -385,6 +389,23 @@ Represents a national ID.
...
+
+<details>
+<summary> StatusObject example </summary>
+<pre>
+{
+    "status": "needs_kyc_data",
+}
+</pre>
+</details>
+
```
LuZhang-Lou, line 408: interestingly, in the travel_rule_data_exchange.md in our private repo, this "preview" dominates but here this is the only(?) occurrence, code block is used elsewhere

PR #24, opened 2020-07-20
LIP2: Libra Roles and Permissions

sblackshear:


lips/lip-2.md
```diff
@@ -0,0 +1,119 @@
...
+---
+This LIP describes the conceptual model and implementation of access control on the Libra blockchain.
+
+---
+# Abstract / Motivation
+---
+
+Libra uses a variant of [role-based access control](https://en.wikipedia.org/wiki/Role-based_access_control) (RBAC) to restrict access to sensitive on-chain operations.
+
+A *role* is an entity with some authority in the Libra system. Every account in the Libra system is created with a single, immutable role that is granted at the time the account is created. Creating an account with a particular role is a privileged operation (e.g., only an account with the ParentVASP role can create an account the ChildVASP role). In some cases, the role is globally unique (e.g., there is only one account with the LibraRoot role). In other cases, there may be many accounts with the given role (e.g., ChildVASP).
```
tzakian, line 20: Nit: (e.g., only an account with the ParentVASP role can create an account _with_ the ChildVASP role)

lips/lip-2.md
```diff
@@ -0,0 +1,119 @@
...
+* MintCurrency(**type**): Create currency of the given **type**
+* BurnCurrency(**type**): Destroy currency of the given **type**
+* PreburnCurrency(**type**): Create a burn request for the currency of the given **type** to be fulfilled or cancelled by the holder of BurnCurrency
+* UpdateExchangeRate(**type**): Update the exchange rate from each non-LBR currency **type** to LBR
+* UpdateDualAttestationLimit: Update the amount above which VASP <-> VASP transactions must use the travel rule protocol
+* {Freeze,Unfreeze}Account: Prevent an account from sending transactions or receiving funds
+* RegisterNewCurrency: Add a new currency type to the system
+* SendWriteSetTransaction: Used to update the Libra protocol by changing on-chain Move modules and/or global state
+* UpdateLibraProtocolVersion: Change the current Libra protocol version. This will atomically update the behavior of node software
+*UpdateVMConfig: Update the transaction script allowlist, module publishing options, or gas costs
```
tzakian, line 96: Nit: Space missing between the * and the `U` is messing up the formatting.

sblackshear:
@junkil-park has had a ton of good feedback about the spreadsheet that was the source material for this LIP and has agreed to take a look tomorrow. Will wait for him and then land!

lips/lip-2.md
```diff
@@ -0,0 +1,119 @@
...
+|ChildVASP(**parent**)                                                                                                                                   |ParentVASP                  |N                                                                                                      |-             |Y                   |Y                                                                                                                                                                                         |Y         |-                                                                                                                                         |
+
+* LibraRoot - The root authority of Libra. Controlled jointly by Libra Payment Networks and the Association Council.
```
dahliamalkhi, line 41: I would leave out organizational and control comments. Instead, say something like:  An entity responsible for creating top-level accounts and performing actions that reconfigure the blockchain.

sblackshear: I think the organizational and control spec is important here. A role is supposed to describe *who* an entity is and permissions are supposed to say *what* an entity can do. Saying "LibraRoot can create top-level accounts and trigger reconfig" is important, but should be expressed in permissions rather than in a description of the role.

lips/lip-2.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+Libra uses a variant of [role-based access control](https://en.wikipedia.org/wiki/Role-based_access_control) (RBAC) to restrict access to sensitive on-chain operations.
+
+A *role* is an entity with some authority in the Libra system. Every account in the Libra system is created with a single, immutable role that is granted at the time the account is created. Creating an account with a particular role is a privileged operation (e.g., only an account with the ParentVASP role can create an account with the ChildVASP role). In some cases, the role is globally unique (e.g., there is only one account with the LibraRoot role). In other cases, there may be many accounts with the given role (e.g., ChildVASP).
+
+A *permission* is the authority to perform a sensitive action in the Libra system. Each permission may be assigned to one or more roles (usually only one), and each role may have zero or more permissions. Permissions can be assigned in genesis, upon account creation (most common), or claimed by an existing account with the appropriate role. Like roles, some permissions are globally unique (e.g., the permission to mint a particular currency type) and some are not.
```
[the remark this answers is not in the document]
sblackshear: I'm not sure I understand the question, but let me try to clarify. Formally, this whole RBAC thing is just two binary relations over three sets:

Sets: Role, Privilege, Account
Relations: Privilege->Role (a many-many relation) and Account->Role (a partial function)

I should probably include this in the LIP.

> Shouldn't it be: a permission for an action (potentially per account) is given to a role **(potentially of a particular account) ?

I think this suggesting that we might also want a Permission -> Account or Role -> Account relations. There's no conceptual problem with this; it's just a more complex model that I don't think we need.

> a role may not be unique: for a particular role, there may be multiple accounts with this role (e.g., ValidatorOperator)

This is expressed by via the Account->Role partial function; there are no restrictions on the range.

> a permission (e.g., UpdateValidatorConfig) is granted to a role (ValidatorOperator)

This is expressed by the Privilege->Role relation

> if there are multiple (Validator) accounts that allow UpdateValidatorConfig, wouldn't a permission to SetValidatorOperator be granted per Validator account to one account with ValidatorOperator role?

This is expressed by parameterizing the permission  (e.g., `RotateAuthenticationKey(addr)` is parameterized by an address). But good  call that I need to parameterize `UpdateValidatorConfig` and `SetValidatorOperator`.

lips/lip-2.md
```diff
@@ -0,0 +1,126 @@
...
+|DesignatedDealer                                                                                                                               |LibraRoot                   |N                                                                                                      |-             |Y                   |N                                                                                                                                                                                         |Y         |2                                                                                                                                         |
```
junkil-park, line 44: According to the current code, DD is granted by TC, which is not consistent to this row.

lips/lip-2.md
```diff
@@ -0,0 +1,126 @@
...
+* SetValidatorOperator(**addr**): Set the address of the operator account that has permission to rotate the consensus public key of the validator account at **addr**
+* RotateDualAttestationInfo(**addr**): Change the public key and endpoint URL that addr uses for the KYC and travel rule protocols
+* RotateAuthenticationKey(**addr**): Rotate the authentication key for the account at **addr**
+* WithdrawCapability(**addr**): Decrease the balance of the account at **addr**
+
+All privileges are granted to an account upon creation, and most remain in that account forever. There are three exceptions: UpdateValidatorConfig, RotateAuthenticationKey and Withdraw.
+* UpdateValidatorConfig is granted to a Validator account upon creation, but can be delegated to a ValidatorOperator. The Validator can subsequently revoke the privilege and delegate to a different operator.
+* Both RotateAuthenticationKey and Withdraw can be be extracted from their original account and placed in a resource elsewhere (including one published under a different account).
+
+### Move Implementation
```
junkil-park, line 116: It would be great if we could somehow connect each element of this policy document to a corresponding formal specification eventually. By the way, arguing that "permission check is implemented correctly" is in-scope of this document?

lips/lip-2.md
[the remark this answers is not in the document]
junkil-park: I also may have some questions or comments on the conceptualization of roles and permissions. I will ponder over it a bit.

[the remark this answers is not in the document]
dahliamalkhi: yes, parameterizing seems to resolve my concern, but not only in the specific cases you mention; also when defined and explained.

lips/lip-2.md
dahliamalkhi: Got it. So let's see if it gets flagged by legal. As a fallback, we could perhaps say: LibraRoot is the root authority of the Libra payment system.

lips/lip-2.md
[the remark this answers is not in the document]
sblackshear: Good call; will explain parameterization up-front.

lips/lip-2.md
sblackshear: Good catch, will fix.

lips/lip-2.md
sblackshear: > It would be great if we could somehow connect each element of this policy document to a corresponding formal specification eventually

Agreed! I think "special cases" requiring side explanation like `UpdateValidatorConfig` and `UpdateValidatorOperator` are indicators of a design that's not as simple as it could be. We could enrich the spec language, but we could (and perhaps should) also simplify the design.

lips/lip-2.md
```diff
@@ -0,0 +1,126 @@
...
+const VALIDATOR_ROLE_ID: u64 = 3;
+const VALIDATOR_OPERATOR_ROLE_ID: u64 = 4;
+const PARENT_VASP_ROLE_ID: u64 = 5;
+const CHILD_VASP_ROLE_ID: u64 = 6;
+```
+
+## Permissions
+The current permissions in Libra are:
+
+|Permission                                                                                                                                          |Granted to                  |Unique?                                                                                                |Transferrable?|
```
junkil-park, line 73: Would it be useful to have an additional column for "granted by"? There are some cases where the code explicitly checks "granters" by assertions (e.g., PreburnCurrency, RotateDualAttenstationInfo, UpdateVMConfig).

Also, would it be unnecessary to have an additional column to specify whether granting of certain permissions is (or should be) done in genesis?

sblackshear: > Would it be useful to have an additional column for "granted by"? There are some cases where the code explicitly checks "granters" by assertions (e.g., PreburnCurrency, RotateDualAttenstationInfo, UpdateVMConfig).

Very good question; I'm not very happy with the way this currently works in the code. It checks "granters" because these permissions are being added at account creation time and granter == creator. What I think we should do eventually is:
1. When creating an account, the first thing we do is generate the role capability for the account to be created
2. Privilege granting code asks for the role capability of the grantee, which the granter can pass down from step 1
This way, we'll match the discipline we discussed offline: privilege granting always checks the role of the grantee.

It actually used to work this way when we had role resources, but falling back to role checks precludes this implementation strategy.

> Also, would it be unnecessary to have an additional column to specify whether granting of certain permissions is (or should be) done in genesis?
Good question; I don't think it would be harmful, but I think it might be overspecification. The way I think of privileges is that *when* they can be granted is irrelevant (and thus need not be specified); all that matters is who can get them.

PR #34, opened 2020-07-31
Add authors fix to individual lip pages

oriooctopus:
This adds a fix similar to the one I did for the LIP metadata table for authors, but this addresses the individual LIP pages

### DIFF
--- a/src/DocComponents/Author/index.js
+++ b/src/DocComponents/Author/index.js
@@ -1,0 +1,20 @@
+import React from 'react';
+
+const Author = ({index, title, username}) => {
+  const formattedTitle = `${index > 0 ? ', ' : ''}${title}`;
+
+  if (!username) {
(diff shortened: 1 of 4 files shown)


oriooctopus:
https://lips.oriooctopus.vercel.app

oriooctopus:
@JoelMarcey  It does not affect that, but yes, multiple linked authors should be allowed. The thing is, if the author entry is something like "Oliver Ullman", we don't take that and assume that the username is @oliverulman. If you want the username to be linked to there are a couple ways:
1. Something like "Oliver Ullman @oriooctopus"
2. Or, and this is the case with the github user libra, you can just put the username as one word. So for example "oriooctopus", or "libra". The code assumes that if the entry is just one word, it is a username.

JoelMarcey:
Ok. That's fair.

PR #35, opened 2020-08-01
LIP-6 : validator configuration management

valerini:
This LIP-6 document covers the validator set configuration management. Ready to review.
One of the remaining questions: should this lip also cover other global on-chain configs? At the moment it only covers the Validator Set on-chain config.


lips/lip-6.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Summary
+---
+This LIP describes the conceptual model and implementation of validator configuration management on the Libra blockchain.
+
+---
+# Abstract / Motivation
+---
+
+The Libra Association is an independent membership organization that is responsible for the governance of the Libra network and the development of the Libra project. Members make key decisions for the Libra network, such as adding a new member, removing an existing member, advancing the version of the Libra protocol, etc. Each member designates a party to run a validator node that operates the blockchain. There is one-to-one relationship between the members and the validator nodes. The validator set defines the current set of validator nodes that run consensus protocol. Validator nodes that are currently part of the set process transactions and interact with each other to reach consensus on the state of the blockchain. Not all of the validator nodes might be in the set at any given time. Members of the Libra Association collectively decide to add or remove validator nodes to or from the validator set.
```
dahliamalkhi: each member designates a party -- I'd use the common term here, a node operator

lips/lip-6.md
```diff
@@ -0,0 +1,85 @@
...
+
+Each validator-owner accounts stores a *consensus configuration* that consists of the consensus public key (to sign consensus messages), the networking information (for the consensus node to connect to other consensus nodes) and the full-node networking information (for the public full-nodes and under certain circumstances for other consensus nodes themselves to connect to this consensus node). The validator-operator designated by the validator-owner is responsible for running an actual consensus node with this consensus config and is responsible for updating/rotating this config as needed.
```
dahliamalkhi, line 32: validator owner account>s<.

lips/lip-6.md
```diff
@@ -0,0 +1,88 @@
...
+
+Each member that participates in operating the blockchain by operating a Validator Node is referred to as a Validator Owner. The Validator Set defines the current set of Validator Nodes that run consensus protocol. Validator Nodes that are currently part of the Validator Set process transactions and interact with each other to reach consensus on the state of the blockchain. The Validator Set may change from time to time and not all members may have Validator Nodes the Validator Set at any given time. Reasons for changing the Validator Set include joining or departure of members from the Association and actions needed for compliance reasons or to comply with LN’s obligations under its payment system license from FINMA to guarantee the secure and performant operation of the platform.
+
```
dahliamalkhi, line 23: Could we consistently use throughout:
Validator Owner (capitals)
Validator Node
Validator Operator

I also wonder whether on-chain, we could change validator accounts to be named validator-owner accounts?

lips/lip-6.md
```diff
@@ -0,0 +1,88 @@
...
+Gradually, the validator owners will start managing their accounts themselves. They will send their new account key to the Association and the Association will rotate their account key to this new key. Right after this rotation transaction gets executed, the validator owner starts to maintain its Libra account.
+
+Both the validator owner account and the validator operator account store a human_name which refers to the name of the company behind the account written in snake case (lowercase, where spaces are replaced with underscores), it is initialized at account creation and never changes. Most validator-related transaction scripts get the address and the names of the account and check that the human_name under the account matches the name passed into the transaction script. This prevents accounts from accidentally mix up the addresses, which are hex strings, and adds a second level of protection against human errors.
+
+### Formal verification guarantees on the Move Implementation
+
+* Disciplined aborts
+* Only LibraRoot or a Validator Operator can trigger reconfiguration
+* Every change to an validator set triggers a reconfiguration
+* Only a validator can set its operator
```
dahliamalkhi, line 71: this contradicts the note above

lips/lip-6.md
```diff
@@ -0,0 +1,88 @@
...
+* Only an operator can update an on-chain config
+* The size of the validator set can only change by at most one when LibraConfig::NewEpochEvent is emitted
+* Multiple calls to add_validator or remove_validator are prohibited except in genesis
+* Both functions register_validator_config or set_validator_config_and_reconfigure check the validity of the consensus key and abort if an invalid key is given
+* An account can only appear once in a validator set
+* (currently) If the validator's operator is not set, the validator owner itself acts as its validator operator
+* If the validator owner changes the operator, its old config continues to persist in the validator set
```
dahliamalkhi, line 78: perhaps replace "continues to persist" with "does not change until explicit remove/add actions on the validator set are triggered.

lips/lip-6.md
valerini: This refers to how the module is written, the fact that LibraRoot is calling it via an admin script (which is essentially a simpler version of the write-set) is a temporary measure. There is always a possibility that the write-set transaction changes the blockchain in some unintended way... What would you suggest writing here?

dahliamalkhi: I see. This makes sense. 

How about : Only Validator Owner (role) can set an operator; for each Validator Owner account, this permission is given to only one Validator Owner (itself).

(as a side note, as mentioned above, an admin script by LibraRoot may act on behalf of Validator Owner accounts.)

lips/lip-6.md
valerini: Great suggestion! I tried to make sure the terminology is consistent and I also added a section Terminology for better clarity.

We could rename on-chain: validator_role to validator_owner_role and rename the script create_validator_account to create_validator_owner account. I added a TODO here: https://github.com/libra/libra/issues/5117

valerini:
It strange, but I cannot add more Reviewers to this pull request. I don't know why, but may be because this pull request now contains your commits, @dahliamalkhi, as well... May be you can add reviewers? I was hoping to add David Dill, David Wolinsky and Sam.

dahliamalkhi:
added them

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Terminology
+* *Validator Node* ("node") is a replica that runs the Libra Protocol to maintain a database of programmable resources. Validator Node processes transactions and interacts with other Validator Nodes to reach consensus on the ordering of transactions and the resulting state of the database after transactions' executions. 
+* *Validator Owner* is a entity that manages an on-chain Libra account of the member of the Libra Association. The account stores the *Validator Config* resource that contains information about the Validator Node of this Validator Owner. This information includes *consensus public key* (EdDSA public key that is used by others to authenticate consensus messages of this node) and *networking information* (network addresses and static Noise keys for establishing secure connections with this node from other nodes or from clients).
```
DavidLDill, line 21: I think it would be better to give Validator Config it's own bullet.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+* *Validator Set* is a copy of Validator Configs of the approved subset of the Validator Owners, it consists of configs currently running the Libra Protocol. The Libra Association collectively decides which Validator Owners' configs are in the set. There is a one-to-one correspondence between the Validator Owners in the Validator Set and the Validator Nodes. The protocol remains safe when at most a third minus one of the Validator Nodes are byzantine.
```
DavidLDill, line 22: I feel like the last sentence should be part of the background before we start the definitions.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+
+Each member that participates in maintaining the blockchain by operating a Validator Node or designating a party to operate a Validator Node is referred to as a Validator Owner. The Validator Set defines the current set of Validator Nodes that run consensus protocol. Validator Nodes that are currently part of the Validator Set process transactions and interact with each other to reach consensus on the state of the blockchain. The Validator Set may change from time to time and not all members may have Validator Nodes in the Validator Set at any given time. Reasons for changing the Validator Set include joining or departure of members from the Association and actions needed for compliance reasons or to comply with LN’s obligations under its payment system license from FINMA to guarantee the secure and performant operation of the platform.
```
DavidLDill, line 17: The transition from the overview of the Association to this paragraph seems a bit abrupt.  Perhaps we could have a paragraph explaining the basic idea of having many validators that reach consensus on a blockchain state.  This would be the place to mention Byzantine faults, etc.  We don't need to go into a lot of detail, but some minimal background about why we need a validator set and the security concerns about who can change it would be helpful (people don't need to know very much about consensus to understand validator sets and configs).

lips/lip-6.md
DavidLDill, line 21: "Validator Owner" is an off-chain concept that is beyond the scope of the Move code.  Enforcing the 1:1 relationship between validator owners and validator nodes is a  rule that is enforced administratively off-chain.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Terminology
+* *Validator Node* ("node") is a replica that runs the Libra Protocol to maintain a database of programmable resources. Validator Node processes transactions and interacts with other Validator Nodes to reach consensus on the ordering of transactions and the resulting state of the database after transactions' executions.
```
DavidLDill, line 20: Should we say that we refer to "validator nodes" as "Validators" in the rest of the doc?

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+* *Validator Operator* ("operator") is an entity that manages its on-chain Libra account and runs a Validator Node on behalf of the Validator Owner who designated this operator to be running the node. There can be one Validator Operator account on-chain that runs two Validator Nodes for two Validator Owners, one node per one owner. It is not allowed to run a single Validator Node per multiple Validator Owners even when the nodes are run by the same Validator Operator. The Validator Operator has the permission to rotate the Validator Configs for its Validator Owners.
+
+# Specification
+## Overview
+Each Validator Owner holds a Libra account with a *Validator Owner role* (see [lip-2: Libra Roles and Permissions](https://github.com/libra/lip/blob/master/lips/lip-2.md)). A Validator Owner designates a Validator Operator to run a Validator Node
```
DavidLDill, line 27: I think we can have consistent and precise language in all of these specifications.  I'm not sure we're all using the same language (we should agree as a group).
Here is how I've been thinking of these concepts in formal specs
1. The Move framework represents Validator Nodes and Validator Operators as addresses.  I'm not sure what terminology others use, but an address that has a published account resource could be called an "account".
2. Every account has a unique role (see roles and permissions)
3. Each Validator account has a Validator role. 
4. Each ValidatorOperator account has a ValidatorOperator role.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
+---
+lip: 6
+title: Libra Validator Configuration Management
+author: Valeria Nikolaenko, David Wolinsky, Sam Blackshear, David Dill
+status: Draft
+type: Informational
+created: 07/30/2020
+---
+# Summary
+This LIP describes the conceptual model and implementation of validator configuration management on the Libra blockchain.
```
DavidLDill, line 10: What is the scope?  When I started reading, I thought it was a specification of the Move Framework:  Suppose someone wanted to implement a validator in C++, what would they have to implement?  But it also includes off-chain concepts and constraints.  Is that to explain the programmatic concepts, or are we also specifying the off-chain rules (e.g., there is exactly one validator node per validator operator?).  (This is a discussion point, not a suggestion for rewriting at this point.)

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+A Validator Owner designates a single Validator Operator account to run a Validator Node. Validator Operator is a Libra account with a *Validator Operator role*. Each Validator Operator may run nodes for multiple Validator Owners, therefore there is a one-to-many relationship between operator accounts and owner accounts. If an operator OP_1 is designated to run Validator Nodes for two Validator Owners V_1 and V_2, the operator must run two distinct instances of Validator Nodes.
+
+Each Validator Owner account stores a consensus configuration that consists of the consensus public key, the networking information (for the Validator Node to connect to other Validator Nodes) and the Full Node networking information (for the public Full Nodes and clients and, under certain circumstances, for other Validator Nodes themselves to connect to this Validator Node). The Validator Operator designated by the Validator Owner is responsible for running a node with this consensus config and is responsible for updating/rotating this config as needed.
```
DavidLDill, line 30: I think we need at least a sentence of background on full nodes in the intro.
"is responsible for" is a little vague.  In the code, ONLY the validator operator may update this information.  In general, we should be precise about who DOESN'T have permission to do something.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+
+The Validator Set is a map of Validator Owner addresses to a copy of their consensus configs. Each Validator Owner address appears in the Validator Set at most once. This set is stored under the LibraRoot address and reflects the configuration that currently runs consensus protocol. Any change to this set defines a new epoch by triggering a new epoch event. If a transaction causes an epoch change, that transaction is the last transaction committed in the current epoch — any subsequent transactions in that block or future blocks from that epoch will be ignored. Once the transaction has been committed, the new Validator Set can start the next epoch of the consensus protocol.
```
DavidLDill, line 32: I think we need a separate discussion of configs.  If someone were trying to re-implement this, it would not be sufficient just to publish the validator set at LibraRoot any way they want (as I understand it) -- clients look at libra configs directly, right?

The discussion of reconfiguration is really important, and probably should appear separate and earlier.

So, yeah, we need a detailed discussion of configs either in this document or elsewhere.  Should that talk about other configs?  Much of the discussion would be about validator config, because it's the hardest.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+
+Only LibraRoot may add/remove Validators Owner accounts to/from the Validator Set. Each Validator Operator may update its validator's config in this set which will also trigger the a new epoch event.
+
+Each Validator Owner is obliged to notify LN about the decision to change the Validator Operator, LN may disallow the switch if the new operator is deemed not trustworthy or the distribution of consensus power becomes skewed towards this operator or operators with similar architectures. LN makes sure there is enough diversity between the Validator Nodes operating the Libra Blockchain.
```
DavidLDill, line 36: This is off-chain administration.  Should we refer to some documentation of policies?

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Move Modules
+
+Four move modules are involved in the validator configuration management
+
+* [`ValidatorConfig`](https://github.com/libra/libra/blob/master/language/stdlib/modules/ValidatorConfig.move): stores local consensus configuration for a validator.
```
DavidLDill, line 42: More precisely, it defines structures (resources?) to hold config information for validators.   It doesn't actually publish this info.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+* [`ValidatorOperatorConfig`](https://github.com/libra/libra/blob/master/language/stdlib/modules/ValidatorOperatorConfig.move): stores the configuration of the validator operator (it only contains a human name).
```
DavidLDill, line 43: Also does not "store" (publish) the info.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+* [`LibraSystem`](): publishes the Validator Set as a global on-chain config and determines the rules for modifying it.
```
DavidLDill, line 44: determines --> enforces?

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+* [`rotate_authentication_key`](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/rotate_authentication_key.move) - to rotate the authentication key of the Validator Owner Account
+
+The following transaction scripts can be run by the Validator Operator:
+* [`register_validator_config`](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/register_validator_config.move) - to register the Validator Config resource
+* [`set_validator_config_and_reconfigure`](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/set_validator_config_and_reconfigure.move) - to change values of the Validator Config resource and update the config in the Validator Set (aborts if the Validator Owner is not in the Validator Set)
+* [`rotate_authentication_key`](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/rotate_authentication_key.move) - to rotate the authentication key of the Validator Operator Account
+
+## Formal verification guarantees on the Move Implementation
```
DavidLDill, line 63: I haven't checked these yet!  This is subject to change when I make progress on that.

lips/lip-6.md
```diff
@@ -0,0 +1,79 @@
...
+* Disciplined aborts
+* Only LibraRoot or a Validator Operator of a Validator Owner which is currently in the Validator Set can trigger reconfiguration
+* Every change to a Validator Set triggers a reconfiguration when LibraConfig::NewEpochEvent is emitted
+* Only a Validator Owner (role) can set its Validator Operator; for each Validator Owner account this permission is given only to the Validator Owner (itself)
+* Only a Validator Operator of a Validator Owner who is present in the Validator Set can update a Validator Config in the Validator Set
+* The size of the Validator Set can only change by at most one in a block
+* Multiple calls to add_validator or remove_validator in the same block are prohibited except in genesis
+* Both functions register_validator_config or set_validator_config_and_reconfigure check the validity of the consensus key and abort if an invalid key is given
+* A Validator Owner account can only appear once in a Validator Set
+* (currently) If the Validator's Operator is not in the Validator Set, the Validator Owner itself acts as its own Validator Operator
```
DavidLDill, line 73: I have a bad feeling about this one.  Why not make the validator owner explicitly be the validator operator, for consistency?

lips/lip-6.md
valerini: Not quite sure I understand your point.
I define the Validator Owner as an on-chain Libra Account, while the Validator Node is a replica (is a server that runs a consensus node). I think the 1:1 relation between the Validator Owner and the Validator Node is enforced by the fact that a single Validator Owner stores a single Validator Config resource that essentially defines the Validator Node (it contains consensus pk, network addresses).

lips/lip-6.md
valerini: Not sure actually, this might become very confusing. I tried to make sure everywhere I use full names ("Validator Node" or "Validator Operator" or "Validator Owner"), though this is a little cumbersome indeed.

lips/lip-6.md
valerini: Great, I was thinking of using this more formal language in the subsection further down, called ## Formal verification guarantees on the Move Implementation. I can add those points there. But I am not quite sure about the terminology, may be we can discuss it over the call.

lips/lip-6.md
valerini: The intention I think was to cover the Validator Set management, where the Validator Set is an on-chain config (one of several we set with LibraConfig::set(...)). Including both on-chain guarantees (with some subset that we could formally verify) and off-chain obligations of each of the parties (the Association should make sure there is enough diversity in the Validator Set, that validator operator should run one node per owner, etc.)

If someone wants to implement the validator in C++ they can refer to Libra Specs (https://github.com/novifinancial/libra-specs).

lips/lip-6.md
valerini: Agree, is responsible is a little vague. But now members are only "contractually or legally responsible" to run nodes. I guess in the future, they will face some financial losses from their deposits if they don't run the nodes or if their nodes are faulty, but at the moment we only enforce this "contractually".

Let me try to add a bit more on full nodes....

lips/lip-6.md
valerini: Good question. @dahliamalkhi, do you know if there are legal docs re-enforcing this?

lips/lip-6.md
valerini: Because there going to be Validator Operator accounts who will manage configs of _multiple_ Validator Owner accounts.

lips/lip-6.md
dahliamalkhi: I agree with Lera: We want Validator Owners representation on chain. We used to (and I believe, still currently) call these accounts simply Validators;  I recommend to use Validator Owner. Albeit slightly longer, it adheres to the language used in the Association legal frameworks, and also disambiguates

PR #45, opened 2020-08-20
Updating LIP5 Wording

dpim:
Rewording descriptions in LIP5 for clarity and to adhere to writing conventions


JoelMarcey:
@dpim Which of these do you want to keep? Then I can resolve the conflict and merge.

### Looking ahead


> In the future, we plan to define additional Account Identifier versions to support other forms of identity, such as more compact subaddress formats. These would leverage a similar overall structure but would have a different version identifier, preventing naming collisions.  

OR 

> In the future, we plan to define additional Account Identifier versions to support other forms of identity, such as a more human-readable subaddress format. These would leverage a similar overall structure but would have a different version identifier, preventing naming collisions.

dpim:
The first one

> In the future, we plan to define additional Account Identifier versions to support other forms of identity, such as more compact subaddress formats. These would leverage a similar overall structure but would have a different version identifier, preventing naming collisions.

PR #54, opened 2020-09-04
Initial draft of LIP-20

tzakian:
This PR is the initial draft for LIP-20. This is meant to provide both definitions for how currencies work on-chain (registration, minting, coin operations, SCS v.s. synthetic currencies, burning), and also some level of intuition at the same time provide some intuition and motivation for each these e.g., `Preburn` exists so that off-chain actions can be performed before burning the coins.

### DIFF
--- a/lips/lip-20.md
+++ b/lips/lip-20.md
@@ -1,0 +1,609 @@
+---
+dip: 20
+title: Diem On-Chain Currency Management
+authors: Tim Zakian, Sam Blackshear
+status: Draft
+type: Informational
+created: 08/28/2020
+---
+
+---
+# Summary
+---
+
+This DIP describes the conceptual model and implementation of currencies on the Diem
+blockchain. Currencies in Diem are represented statically at the type level, each
+currency with a unique Move type. A type is treated as a currency in Diem
+if and only if it has previously been registered as a currency on-chain. Once a currency
+is registered it remains registered forever.
+
+Every currency that is registered on-chain has metadata
+associated with it at the time of registration. That metadata includes the currency code, value
+held on-chain in the currency, and other relevant information[[1]](#registration_and_metadata). One of the most
+important properties held in a currency's metadata is whether it is a Single Currency Stablecoin
+(SCS), i.e., a fiat currency, or a synthetic currency that may consist of one or
+multiple other currencies registered on-chain, e.g., ≋XDX.
+
+On-chain, all assets with a value of `x` in a registered currency `C` are held as a "Diem Coin" resource
+of type [`Diem<C>`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#resource-diem)
+with its `value` field set to `x`. Every Diem Coin has a certain set of
+operations that may be performed on it and other coins of the same
+currency[[2]](#general_currency_operations), and every registered currency possesses a
+specific set of metadata about itself[[3]](#metadata_spec) that may be
+queried[[4]](#currency_ops). All operations must be performed on a registered
+currency type, with the exception of registration.
+
+---
+# <a name="requirements">Requirements</a>
+---
+
+This DIP describes the definition of currencies on-chain and their registration
+process (in a technical setting), the representation of assets in a given
+currency, and the operations that may be performed with registered currencies
+and assets in a registered currency (or "Diem Coins").
+
+The requirements for a currency in Diem are as follows:
+1. *Privileged Registration*: The registration of a Move type as a currency
+   on-chain is restricted to accounts with the
+   [`RegisterNewCurrency` permission](https://github.com/libra/lip/blob/master/lips/lip-2.md#permissions).
+2. *Privileged minting and burning*: Every on-chain currency must have a way
+   for new coins to minted and burned from the system. Additionally, minting
+   and burning of Diem Coins of any currency type must be a privileged action
+   and tightly controlled yet also flexible in its specification for each
+   currency that is defined. This is so that it can be used, e.g., to grant
+   specific designated addresses the ability to mint coins of a specific
+   currency, or to ensure that certain preconditions are met before coins are
+   minted or burned.
+3. *Explicit representation*: The representation of value in a currency is
+   explicitly represented at the type level in Move. Move values that have a
+   Diem Coin type of a given currency cannot be accidentally or purposefully
+   misinterpreted as anything other than a coin in the given value and
+   currency.
+4. *Fungibility and correctness*: The way currencies may interact and be
+   interchanged must always be well-defined. Any undefined or invalid behaviors
+   such as trying to combine coins of different currency types must never be
+   allowed, and error conditions well specified.
+5. *Conservation, and encapsulation of value*: The value of a coin in
+   circulation can never be accidentally or purposefully altered except through
+   the specified operations defined in this document. Additionally, the
+   `total_value` of all coins minted in a currency must equal the sum of the
+   values of all coins in circulation in that currency no matter the operations
+   performed, with the exception of minting and burning operations.
+6. *Value normalization*: Every currency must have a way to _roughly_ normalize
+   its value to a specified currency so that values expressed across different
+   currencies may be compared for use in the system. In Diem, this specified
+   currency is ≋XDX.
+
+The Diem currency management on-chain is designed to provide a common interface that can
+be used for any currency that ensures that these key properties are kept.
+
+---
+# <a name="registration_and_metadata">Registration of Currencies and Currency Metadata</a>
+---
+
+Every currency on-chain is represented as a Move-defined type `C`. This type can be
+either a resource or struct type. In order for the system to view a type `C` as
+representing a currency on-chain, it must first be _registered_ as such. This may be done by
+calling one of either [`Diem::register_SCS_currency`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-register_scs_currency)
+or [`Diem::register_currency`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-register_currency)
+instantiated with the type `C` that is to be registered as a new SCS or synthetic currency respectively.
+
+## <a name="metadata_spec">Required information for registration of a currency and metadata</a>
+
+When a currency is registered using these functions, a number of metadata need to be
+provided about the currency that is being registered. In particular, every registered
+currency on-chain contains the following data, names in bold are metadata that must be
+provided at the time of registration:
+
+| Name                          | Type                                   | Mutable      | Description                                                                                                                                      |
+| ----------------------------- | -------------------------------------- | ------------ | -----------------------------------------------------------------------------------------------------------------------                          |
+| **to_xdx_exchange_rate**      | FixedPoint32                           | true         | The _rough_ exchange rate from `C` to XDX. Used only for dual attestation threshold and mempool gas price normalization                          |
+| **is_synthetic**              | bool                                   | false        | Whether `C` is a synthetic currency or not (if false, then it is an SCS currency). Informational only with the except for its effects on events. |
+| **scaling_factor**            | u64                                    | false        | The scaling factor that a coin's value in `C` should be multiplied by to arrive at the off-chain "real world" value                              |
+| **fractional_part**           | u64                                    | false        | The smallest fractional part (number of decimal places) to be used in the human-readable representation of the currency                          |
+| **currency_code**             | vector\<u8\>                           | false        | The currency code (e.g., "XDX") for `C`                                                                                                          |
+| total_value                   | u128                                   | true         | The total value of all currency in circulation on-chain, initialized to zero                                                                     |
+| preburn_value                 | u64                                    | true         | The total value of all coins in this currency waiting to be burned                                                                               |
+| can_mint                      | bool                                   | true         | Whether more value in this currency can be added to circulation                                                                                  |
+
+All of the metadata about a specific currency `C` is held as part of the
+[`CurrencyInfo<C>`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#resource-currencyinfo)
+resource that is uniquely published for each registered currency type `C` under the [Diem Root](https://github.com/libra/lip/blob/master/lips/lip-2.md#roles) (see also the implementation of
+[`Diem::register_SCS_currency`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-register_scs_currency)
+and
+[`Diem::register_currency`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-register_currency))
+account address. We now describe each of these data and their purpose.
+
+### <a name="to_xdx_exchange_rate">`to_xdx_exchange_rate`</a>
+
+There are certain cases where values in two different currencies need to be
+compared against each other. This exchange rate is used to normalize both values to their
+approximate ≋XDX value so that their values may be compared. In particular:
+* Transactions can set their gas price in different currencies, however transactions
+  need to be ranked against each other based upon their gas prices. This field is used by
+  the rest of the system to normalize gas prices to one single "unit" for comparison and
+  ranking purposes.
+* The dual attestation threshold needs to apply to payments in all currencies. This limit is set in
+  terms of ≋XDX, transfers are first normalized using this exchange rate and then compared
+  against the limit.
+
+#### Mutability
+The initial exchange rate to ≋XDX must be provided at the time of registration. The
+exchange rate may be [updated](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-update_xdx_exchange_rate)
+by an account with the [Treasury Compliance role](https://github.com/libra/lip/blob/master/lips/lip-2.md#roles) to correspond to
+fluctuations in the real-time exchange rate between the specified currency `C` and
+≋XDX. Note that this is not meant to be an exact exchange rate and _should not_ be used for
+determining exchange rates for value transfer between different currencies.
+
+### `is_synthetic`
+
+This defines whether the given currency is synthetic or an SCS. In the case of
+[`Diem::register_SCS_currency`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-register_scs_currency)
+the function sets this to `false`. For
+[`Diem::register_currency`](https://github.com/libra/diem/blob/master/language/stdlib/modules/doc/Diem.md#function-register_currency)
+this must be provided to the function.
+
+#### Mutability
+Immutable
+
+### <a name="scaling_factor">`scaling_factor`</a>
+
+This defines what fraction each unit of value on-chain represents in the
+off-chain representation of the currency. As an example, if the on-chain
+unit of value in a currency `C` is one millionth of `C`'s smallest real-world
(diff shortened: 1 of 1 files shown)


lips/lip-20.md
```diff
@@ -0,0 +1,425 @@
...
+register a currency with an already taken currency code will fail. There are no
+restrictions on the format or length of currency codes, but there are
+[operational restrictions](#operational_things) that should be considered from the
+on-chain perspective when choosing a currency code.
+
+---
+# <a name="coin_value_representation">Representation of Value</a>
+---
+
+Every currency that is registered on-chain holds a [`scaling_factor`](#scaling_factor)
```
dahliamalkhi, line 229: This is already mentioned when scaling_factor is introduced. Remove?

lips/lip-20.md
```diff
@@ -0,0 +1,425 @@
...
+
+### Predicates
+* [`Libra::is_currency<CoinType>(): bool`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-is_currency)
+    - Related assertion: [`Libra::assert_is_currency<CoinType>()`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-assert_is_currency)
+* [`Libra::is_SCS_currency<CoinType>(): bool`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-is_SCS_currency)
+    - Related assertion: [`Libra::assert_is_SCS_currency<CoinType>()`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-assert_is_SCS_currency)
+* [`Libra::is_synthetic_currency<CoinType>(): bool`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-is_synthetic_currency)
+
+---
+# <a name="off_chain_use">Off-Chain Interactions</a>
```
dahliamalkhi, line 403: Why is this titled Off-Chain?

dahliamalkhi (commented):
hurray for getting this started!!! see comments inline.

lips/lip-20.md
tzakian: This section is really talking about the events emitted for currencies and when they're emitted, so I think this isn't a good title for this section. I'll retitle the section to clarify this.

sblackshear (commented):
Looks like a fantastic start--thanks for writing this!

The biggest question I have is whether we want this to be a Libra-specific standard (in which case I think it is close to ready) or a more general standard that will be adopted by other blockchains that want to interop with Libra and/or use Move (in which case I think we have some thinking to do in a few key areas). Some of my comments are assuming we want the latter (especially given the "20"), but wasn't sure.

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+---
+
+This LIP details the representation of currencies on-chain, their
+registration process (in a technical setting), the representation of assets in a specific
+currency, and the operations that may be performed with registered currencies and currency
+assets (or "Libra Coins").
+
+The central requirements for a currency representation in Libra can be
+distilled to a couple core requirements:
+1. *Creation and removal*: Every on-chain currency must have a way for new
```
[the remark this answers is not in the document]
tzakian: Good call out. I'll update across the doc.

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+   coins to created and removed from the system.
+2. *Privilege and control*: Creation and removal of Libra Coins of any currency
+   type must be a privileged action and tightly controlled, as every coin in
```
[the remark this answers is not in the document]
tzakian: Good point. I'll update to remove reserve and update with "controlled yet flexible" here.

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+### <a name="scaling_factor">`scaling_factor`</a>
+
+This defines what fraction each unit of value on-chain represents in the off-chain
+representation of the currency. As an example: the on-chain representation of ≋LBR is
+microlibra, therefore each unit of value on-chain represents one microlibra and the
+`scaling_factor` for ≋LBR is `1,000,000` since each unit is `1/1,000,000` of an ≋LBR.
+
+### `fractional_part`
+
+This defines the truncation that needs to be performed when transferring from the on-chain
```
bob-wilson, line 149: "Truncation" doesn't seem like the right word for this. A client may choose to display smaller fractions, as is often done for things like stock prices. I like the phrase here "smallest denomination real-world coin". Maybe this could be phrased in terms of that, with a suggestion that values be rounded (probably not truncated) for display in those units?

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+the restrictions of the particular currency. For example, if an SCS `C` has a scaling factor of
+`1,000,000` but the smallest denomination real-world coin for `C` is 100'th of a `C`, then
+the `fractional_part` for `C` would be `100`.
+
+#### Mutability
+Immutable
+
+### `currency_code`
+
+The currency code is specified at the time of registration of the currency. In addition to
```
bob-wilson, line 160: Should this reference ISO 4217? It's important that currency codes be unique, but at least at first, I'm assuming that our currencies will be unofficial.

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+[`Libra::register_SCS_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-register_scs_currency)
+function, both of these capabilities are stored under the Treasury Compliance account at address
+`0xB1E55ED`. In the case that the [`Libra::register_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-register_currency)
+is used to register the currency both of these capabilities are returned to the caller, which may store them in other structures.
+The `MintCapability` and `BurnCapability` for each currency are [unique](https://github.com/libra/lip/blob/master/lips/lip-2.md#permissions); after
+registration of a currency no future mint and burn capabilities for the currency may be created.
+
+## On-chain list of registered currencies
+
+Whenever a currency is registered on-chain, the to-be-registered currencies currency code is added to the
```
bob-wilson, line 219:
```suggestion
Whenever a currency is registered on chain, its currency code is added to the
```

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+[`RegisteredCurrencies` on-chain config](https://github.com/libra/lip/blob/master/lips/lip-6.md#registered-currencies).
+The set of currency codes registered on-chain is always a set. Attempting to
```
bob-wilson, line 221: "the set... is always a set": there must be a better way to say that!

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+register a currency with an already taken currency code will fail. There are no
+restrictions on the format or length of currency codes, but there are
```
bob-wilson, line 223: There are indeed some restrictions. Currently they are required to be Move identifiers, and I just proposed a PR to further require that they be composed solely of alphanumeric characters (https://github.com/libra/libra/pull/6081) with a follow-up issue to see if we can restrict use of lowercase characters (https://github.com/libra/libra/issues/6082)

bob-wilson (commented):
I didn't get through all the details yet, but here are some comments, including some very nit-picky punctuation stuff

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+# Summary
+---
+
+This LIP describes the conceptual model and implementation of currencies on the Libra
+blockchain. Currencies in Libra are represented statically at the type level, each
+currency with a unique Move type. A type may treated as a currency in Libra
+if and only if it has previously been registered as a currency on-chain. Once a currency
+is registered it remains registered forever.
+
+Every currency that is registered on-chain has metadata
```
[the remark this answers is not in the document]
tzakian: > There should be commas both before and after "i.e." and "e.g." -- you've only got commas after those. It helps if you think of spelling them out as "that is" or "for example".

Good call. I'll update :) 

> I can see why you were tempted to use a semicolon between "a SCS...fiat currency" and "or a synthetic currency..." because there are already lots of commas, but I don't think a semicolon works there. It should be a comma.

Yea...I might look in to rephrasing that. Not having a clean separation there seems not very nice.

> You should have something to separate "metadata associated with it at the time of registration" from "such as the currency-code....". A comma would work, or else you could make that a separate sentence: e.g., "That metadata includes..."

👍 

> Should "currency-code" be hyphenated? I think not.

👍 

> How about "on-chain"? In this case, it depends on the usage (and I know that this is not done consistently, but hey, I'm already writing all this about punctuation!) In a phrase like "currencies registered on chain", I don't believe a hyphen should be used. The "on chain" is a prepositional phrase described where the currencies are registered. If you refer to something like "on-chain resources", the "on-chain" is used as an adjective describing the resources. (There's surely a more precise term for this, but I forget what it is.)

Ahh! yes good point. The rule of thumb I use was given to me by a PL person: use "-" for words that need to be bound to the left... So not nearly as precise as yours 😉 

> ...and then there's the Oxford comma question (before "and other information"). I tend to use those myself and since you're an Oxford grad, I would have expected one here 

They actually just called. They're planning on revoking my degree because of this.... In all seriousness thanks for pointing it out since I am a very big fan (and user!) of the Oxford comma. I'll fix :)

lips/lip-20.md
tzakian: Good call on the truncation terminology and using "smallest denomination real-world coin" instead. I'll update and include the suggestion to round and not truncate as well.

lips/lip-20.md
tzakian: Good question. I purposefully avoided being too precise here since I don't think the exact format is locked down on this yet. This also might get a bit to Sam's high-level question at the top of this PR about how Libra-implementation-specific we want this to be. My thinking for now it to leave this intentionally vague for now and circle back if we want to specify this more heavily?

lips/lip-20.md
tzakian: It can't be a false statement though! :p 

I'll update.

lips/lip-20.md
tzakian: I'll update to include those changes!

tzakian:
> The biggest question I have is whether we want this to be a Libra-specific standard (in which case I think it is close to ready) or a more general standard that will be adopted by other blockchains that want to interop with Libra and/or use Move (in which case I think we have some thinking to do in a few key areas). Some of my comments are assuming we want the latter (especially given the "20"), but wasn't sure.

My thinking was that this would be more Libra specific, but I defer to your and @dahliamalkhi's view on this, and am happy to edit this either direction.

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+## <a name="metadata_spec">Required information for registration of a currency and metadata</a>
+
+When a currency is registered using these functions, a number of metadata need to be
+provided about the currency that is being registered. In particular, every registered
+currency on-chain contains the following data, names in bold are metadata that must be
+provided at the time of registration:
+
+| Name                          | Type                                   | Mutable      | Description                                                                                                             |
+| ----------------------------- | -------------------------------------- | ------------ | ----------------------------------------------------------------------------------------------------------------------- |
+| **to_lbr_exchange_rate**      | FixedPoint32                           | true         | The _rough_ exchange rate from `C` to LBR. Used only for dual attestation threshold and mempool gas price normalization   |
```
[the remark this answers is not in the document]
tzakian: I've added a comment to the requirements portion to mention that it needs to normalize to a selected currency/value and in Libra this is chosen to be LBR.

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+
+All of the metadata about a specific currency `C` is held as part of the
+[`CurrencyInfo<C>`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#resource-currencyinfo)
+resource that is uniquely published for each registered currency type `C` under the [Libra Root](https://github.com/libra/lip/blob/master/lips/lip-2.md#roles) (see also the implementation of
+[`Libra::register_SCS_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-register_scs_currency)
+and
+[`Libra::register_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-register_currency))
+account address. We now describe each of these data and their purpose.
+
+### `to_lbr_exchange_rate`
```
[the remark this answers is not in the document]
tzakian: I went for a table here since the `CurrencyInfo` struct contains a number of other fields related to events that I don't want to go into at this point in the document. I'm open to changing this to the full resource definition, but I do have a preference for the table though because of this. But, like I said, I'm open to changing it if you still think it should be the resource definition after this :)

lips/lip-20.md
```diff
@@ -0,0 +1,426 @@
...
+
+It is important to note that in the system any value of type `Libra<Currency>` is considered a
+valid asset in the `Currency` currency with its value given by the `value` field.  Therefore the
+creation of any non-zero value of this type needs to be considered a privileged operation
+since it represents creation of value on-chain. Additionally, this is an if-and-only-if
+relation; a value represents a specific amount `v` in a given currency `Currency`, if, and only
+if the value has the type `Libra<Currency>`.
+
+---
+# <a name="minting_and_burning">Creation and Removal of Value</a>
```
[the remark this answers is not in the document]
tzakian: Agreed that adding code in will be helpful here. But, I think it's important to not go to code heavy, and I think walking a middle ground here would be beneficial. Particularly, in my thinking, since this is an informational LIP I don't want to describe "this is how it's done" but also to some extent "this is why it's done."

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+(SCS), i.e., a fiat currency, or a synthetic currency that may consist of one or
+multiple other currencies registered on-chain, e.g., ≋LBR.
+
+On-chain, all assets with a value of `x` in a registered currency `C` are held as a "Libra Coin" resource
+of type [`Libra<C>`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#resource-libra)
+with its `value` field set to `x`. Every Libra Coin has a certain set of
+operations that may be performed on it and other coins of the same
+currency[[2]](#general_currency_operations), and every registered currency possesses a
+specific set of metadata about itself[[3]](#metadata_spec) that may be
+queried[[4]](#currency_ops). Additionally, none of the operations detailed in the rest of
```
sblackshear, line 33: This sentence has too many negatives to grok easily :). Perhaps: "All operations must be be performed on a registered currency type"?

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+this document may be performed on a non-registered currency (i.e., type) with the
+exception of registration.
+
+---
+# <a name="requirements">Requirements</a>
+---
+
+This LIP details the representation of currencies on-chain, their
```
sblackshear, line 41: "representation of currencies" and "representation of assets in a specific currency" sound similar--maybe different phrasing?

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+registration process (in a technical setting), the representation of assets in a specific
+currency, and the operations that may be performed with registered currencies and currency
+assets (or "Libra Coins").
+
+The requirements for a currency in Libra are as follows:
+1. *Privileged Registration*: The registration of a Move type as a currency
+   on-chain must be a privileged operation, and only possible to accounts that have the
```
sblackshear, line 48:
```suggestion
   on-chain is restricted to accounts with the
```

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+## <a name="metadata_spec">Required information for registration of a currency and metadata</a>
+
+When a currency is registered using these functions, a number of metadata need to be
+provided about the currency that is being registered. In particular, every registered
+currency on-chain contains the following data, names in bold are metadata that must be
+provided at the time of registration:
+
+| Name                          | Type                                   | Mutable      | Description                                                                                                                                      |
+| ----------------------------- | -------------------------------------- | ------------ | -----------------------------------------------------------------------------------------------------------------------                          |
+| **to_lbr_exchange_rate**      | FixedPoint32                           | true         | The _rough_ exchange rate from `C` to LBR. Used only for dual attestation threshold and mempool gas price normalization                          |
```
sblackshear, line 101: Will we still have this at launch?

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+
+#### Mutability
+Immutable
+
+### <a name="scaling_factor">`scaling_factor`</a>
+
+This defines what fraction each unit of value on-chain represents in the off-chain
+representation of the currency. As an example: the on-chain representation of ≋LBR is
+microlibra, therefore each unit of value on-chain represents one microlibra and the
+`scaling_factor` for ≋LBR is `1,000,000` since each unit is `1/1,000,000` of an ≋LBR.
```
sblackshear, line 155: This example is extremely helpful! But would be good to use the same dummy `C` type for consistency with the `fractional_part` example.

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+#### Reference Implementations
+
+* [`Libra::is_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-is_currency)
+    - Related assertion [`Libra::assert_is_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-assert_is_currency)
+* [`Libra::is_SCS_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-is_SCS_currency)
+    - Related assertion [`Libra::assert_is_SCS_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-assert_is_SCS_currency)
+* [`Libra::is_synthetic_currency`](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/Libra.md#function-is_synthetic_currency)
+
+---
+# <a name="currency_events">Currency-Related Events</a>
```
sblackshear, line 568: If it's not too much work, could be useful to discuss reconciliation of events here. E.g., 

sum(values in all mint events for C) - sum(values in all burn events for C) = total value for C

lips/lip-20.md
```diff
@@ -0,0 +1,595 @@
...
+
+When deciding the on-chain currency code for a currency,
+the module, as well as the type name chosen for the currency must match the on-chain currency
+code chosen for the currency. e.g., if you want to use currency code `"ABC"` for a
+currency, then the fully qualified Move type that is used for that currency on-chain
+_must_ be `0x1::ABC::ABC`.
+
+Because of this, and in addition:
+* An on-chain currency code must be a valid Move identifier; and
+* It can consist solely of alphanumeric (TODO: [uppercase](https://github.com/libra/libra/issues/6082)) characters.
```
sblackshear, line 595: Are we planning to resolve this TODO when we do the coin rename?

sblackshear (approved):
Nice!

lips/lip-20.md
tzakian: yep! We keep this around so that we won't need to upgrade the `CurrencyInfo` resource, and because this is used for both the dual attestation normalization and gas price normalization.

lips/lip-20.md
tzakian: yep! Although I think I'll remove the TODO since it isn't really adding anything IMO

tzakian:
I believe all comments have been resolved. I've also updated to use Diem and XDX everywhere, with the exception of links at the moment. 

@sblackshear let me know if this looks good to you.

PR #66, opened 2020-09-22
LIP-6: More on specifications, corrections/clarifications through the text

valerini:
* Adding more clarifications
* Better flow for the document
* Fix the notations (to use capital terms everywhere)
* Add a picture to lift confusion
* Adding more concrete things on specifications


lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+<img alt="Validator Set" src={useBaseUrl('img/validator_set.svg')} />
+
+# Specifications
+
+## Workflows
+### Validator Owner and Validator Operator accounts creation
+Only the Association (through the Association Operation Services) may create Validator Owner and Validator Operator accounts.
+Validator Owners and Validator Operators should contact the Association off-chain and communicate their authentication keys as well as the names of their organizations written in snake case (lowercase, where spaces are replaced with underscores). The authentication key is a 32 elements byte array, derived from the public verification key. The protocol currently supports Ed25519 and MultiEd25519 signature schemes.
```
davidiw, line 36: not in this revision, but we need to be clear that some accounts are self-managed and eventually all will be but some are also managed by the association on behalf of owners. these accounts have an auth key of all 0 and never directly submit transactions to the chain but instead are controlled via libra root ... or something like that.

valerini: There is a special Note section at the end of the doc on that.

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+status: Draft
+type: Informational
+created: 07/30/2020
+---
+# Summary
+This LIP describes the conceptual governance model of the Libra blockchain, it discusses the management of its main component - the on-chain validator set configuration.
+Validator set is part of a global on-chain configuration that additionally includes the list of registered currencies, publishing policies and Libra version.
+
+# Abstract / Overview
+The Libra Payment Network (“LPN”) is a blockchain-backed payment platform and an infrastructure for stablecoins. Both the blockchain and the stablecoin financial instruments are controlled by Libra Networks (“Libra Networks” or “LN”) a subsidiary of the Libra Association (the “Association” or “LA”). Libra Networks is in the process of applying for a payment system license from the Swiss Financial Market Supervisory Authority (“FINMA”). Once granted, LN will be responsible for ensuring that LPN operates in compliance with the payment system license from FINMA (the "FINMA License").
```
davidiw, line 14: Let's say LibraRoot which is a shared entity on chain between LA and LPN and use LibraRoot instead of association hereafter

lips/lip-6.mdx
dahliamalkhi: seems to be addressed in the Notes section at the bottom?

dahliamalkhi (changes requested):

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+Each member of the Libra Association is expected to be ready to run a replica of the Libra protocol, called a Validator Node. Validator Nodes coordinate and reach consensus on the state of the blockchain by processing transactions, they ensure correct updates even if some Validator Nodes (fewer than a third) fail in arbitrary ways. Though only a subset of members, called a Validator Set, run Validator Nodes in any given time. The Association determines the composition of the Validator Set by adding and removing members over time. Reasons for changing the Validator Set include new members joining the Association or existing members leaving it and actions needed to comply with LN’s obligations under its FINMA License designed to guarantee the secure and performant operation of the platform. The Validator Set is stored as part of the blockchain state, in an on-chain configuration. 
+
+# Terminology
+* *Validator Owner* is an owner of the Libra account of the Libra Association's member. The account stores the *Validator Config* resource.
```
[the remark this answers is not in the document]
davidiw: or more abstract as ... is an account capable of storing a Validator Config which defines an end point capable of participating  in Libra's proof of authority consensus protocol

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* *Validator Config* is an on-chain resource that contains information for running a Validator Node. It includes a *consensus public key*, which is an EdDSA public key used to authenticate consensus messages. It also includes *network information*: network addresses and static public keys for establishing secure connections with this Validator Node from other Validator Nodes or from clients.
```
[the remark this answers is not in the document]
davidiw: we could probably refer to the network specification and leave network addresses as ambiguous: https://github.com/libra/libra/blob/master/specifications/network/network-address.md

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* *Validator Set* is an on-chain resource that stores a vector, where each entry contains an address of a Validator Owner and a copy of its Validator Config. Though the Validator Config in the Validator Set may be an older version of the Validator Config currently stored under Validator Owner's account, each Validator Node is run with the Validator Config from the Validator Set.
```
davidiw, line 23: Validator set defines the current set of validators performing the consensus protocol. Specifically it is a vector containing validator config entries that map to the validators current states. These entries may differ...

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+Each address appears at most once in the vector. The Libra Association collectively decides which Validator Owners' addresses are in the Validator Set. There is a one-to-one correspondence between a Validator Owner in the Validator Set and a Validator Node. The Validator Set changes, either when one of the Validator Nodes updates their parameters or when a new member is added or removed, the Reconfiguration happens.
```
davidiw, line 24: We should probably re-read and unify validator config and validator node where appropriate and be clear that a validator config defines either the current or potential configuration for a validator node.

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* *Validator Node* is a replica that runs the Libra Protocol to maintain the blockchain state. A Validator Node processes transactions and interacts with other Validator Nodes to reach consensus on the ordering of transactions and the resulting state of the database after transactions' executions.
```
davidiw, line 25: maybe consider the following order:

validator node defines a participant in consensus
validator config defines identity
validator owner is responsible for this
validator operator manages this
validator set is the current participants
finally reconfiguration

lips/lip-6.mdx
davidiw: that is technically true but I prefer to see it unified so it doesn't feel so ... much as an after thought as this special note applies to /every/ validator owner in genesis.

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+```
+It contains a `human_name` of the account which is set at the account creation time and cannot be changed. The `config` and `operator_account` fields are initially set to `none`. The `config` is a place holder for parameters of the Validator Node.
+```
+struct Config {
+  consensus_pubkey: vector<u8>,
+  validator_network_addresses: vector<u8>,
+  fullnode_network_addresses: vector<u8>,
+}
+```
+
```
davidiw, line 55: we should evaluate if we want these fields to be lcs encoded... as of now ... they are technically in a grey area. cc: @zekun000

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* *Validator Operator* is an owner of an on-chain Libra account, it may operate none, one or several Validator Nodes: a Validator Owner may designate a Validator Operator to run a Validator Node on Validator Owner's behalf. A Validator Operator may run multiple Validator Nodes. The Validator Operator for a Validator Owner has permission to rotate the Validator Config for the Validator Owner locally, as well as copy it to the Validator Set.
+* *Reconfiguration* defines a new epoch by triggerring a new epoch event (`LibraConfig::NewEpochEvent`), it gets emitted if any part of the global on-chain configuration, such as the Validator Set, changes. If a transaction causes an epoch change, that transaction is the last transaction committed in the current epoch — any subsequent transactions in that block or future blocks from that epoch will be ignored. Once the transaction has been committed, the new configuration starts being valid immediately. E.g. the new Validator Set will start running the consensus protocol in the new epoch.
+
+<img alt="Validator Set" src={useBaseUrl('img/validator_set.svg')} />
+
+# Specifications
+
+## Workflows
```
davidiw, line 33: I think I had some feedback here to break this up into two sections:
* a set of operations
* a set of workflows that give high level overview and dictate you through a series of operations

not sure the best ordering

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+## Move Modules
+
+Four move modules are involved in the validator configuration management
+
+* [`ValidatorConfig`](https://github.com/libra/libra/blob/master/language/stdlib/modules/ValidatorConfig.move): defines local consensus configuration for a Validator Node of a Validator Owner.
+* [`ValidatorOperatorConfig`](https://github.com/libra/libra/blob/master/language/stdlib/modules/ValidatorOperatorConfig.move): defines the configuration of the Validator Operator (it only contains a human name).
+* [`LibraSystem`](https://github.com/libra/libra/blob/master/language/stdlib/modules/LibraSystem.move): publishes the Validator Set as a global on-chain config and enforces the rules for modifying it.
+* [`LibraConfig`](https://github.com/libra/libra/blob/master/language/stdlib/modules/LibraConfig.move): a generic module that manages all of the global on-chain configurations, including the Validator Set configurations.
+
+## Transaction Scripts
```
davidiw, line 104: do we want to talk about the nonce scripts or admin scripts? move currently doesn't allow setting an auth key to all zeroes when creating an account, so it has to be done with the admin script

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+## Formal verification guarantees on the Move Implementation
+TODO: These may change as verification work proceeds.
+
+* The Move framework represents Validator Nodes and Validator Operators as addresses.  Each such address has a published account.
+* Every account has a unique role (see roles and permissions)
+* Each Validator account has a Validator role.
+* Each ValidatorOperator account has a ValidatorOperator role.
+* Only LibraRoot or a Validator Operator of a Validator Owner which is currently in the Validator Set can trigger reconfiguration
+* Every change to a Validator Set triggers a reconfiguration when LibraConfig::NewEpochEvent is emitted
+* Only a Validator Owner (role) can set its Validator Operator; for each Validator Owner account this permission is given only to the Validator Owner (itself).
```
davidiw, line 129: via admin scripts this is not true, LibraRoot can as well, how are we thinking about this quirk?

davidiw: cc: @DavidLDill

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* Only a Validator Operator of a Validator Owner who is present in the Validator Set can update a Validator Config in the Validator Set
```
davidiw, line 130: * Only LibraRoot or a Validator Operator of a Validator Owner which is currently in the Validator Set can trigger reconfiguration

these two points overlap a lot, curious why they are separate...

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* The size of the Validator Set can only change by at most one in a block
+* Multiple calls to add_validator or remove_validator in the same block are prohibited except in genesis
```
davidiw, line 132: do we want to be explicit that these apply only to move and writesets can violate these properties or should we use the term writesets? cc: @zekun000 nomenclature!

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* Both functions register_validator_config or set_validator_config_and_reconfigure check the validity of the consensus key and abort if an invalid key is given
+* A Validator Owner account can only appear once in a Validator Set
+* (currently) If the Validator's Operator is not in the Validator Set, the Validator Owner itself acts as its own Validator Operator
```
davidiw, line 135: oh, we never fixed this :O, how hard is it to turn this off? I think all uses cases in code are two distinct accounts

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* If the Validator Owner changes the Validator Operator, its old Validator Config does not change in the Validator Set until a new Validator Operator sets a new Validator Config and explicit calls an update (LibraSystem::update_config_and_reconfigure) or the Association calls add/remove (LibraSystem::add_validator/LibraSystem::remove_validator).
```
davidiw, line 136: this is a bit confusing as it contradicts the above... if I am an owner acting as my own operator, why can I not reconfigrue?

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* Disciplined aborts
+
+# Notes
+In the first version of the Libra Protocol, Validator Owners will not maintain ownership of their Libra accounts; rather the LibraRoot account will run Validator Owners' operations on their behalf. Validator Operators on the other hand will maintain full ownership of their Libra accounts. Gradually, the Validator Owners will start managing their accounts themselves. A Validator Owner will send their new account key to the Association, the Association will rotate their account key to this new key and right after this rotation transaction gets executed, the Validator Owner starts to maintain its Libra account.
```
davidiw, line 140: I'd prefer for us to release a new lip that updates our governance policy and strike this when appropriate rather than try to accommodate versioning now.

davidiw (commented):
This is a truly amazing piece of work, I know I gave a lot of feedback, but that is just indicative of how important and amazing I think this work is :D

lips/lip-6.mdx
zekun000: I don't think this is precise, multiple calls in a single txn or multiple add|remove_validator txn?

genesis is not a block, it's a single writeset txn. writeset doesn't call those functions, it's just a combined result of multiple calls, if we want to discuss nomenclature lol

davidiw: Great point, how does the formal verification prove this as this is a property of the executor afaiu. Should this doc even talk about blocks?

lips/lip-6.mdx
valerini: This is not correctly written - "If the Validator Owner did not set the Validator Operator, the Validator Owner itself acts as its own Validator Operator" - this should be here instead.

lips/lip-6.mdx
valerini: Only the operator can reconfigure to avoid race conditions. The Owner is its own Operator if the Owner didn't set the Operator. So if the Operator is unset, the owner = operator, so the owner can reconfigure.
Perhaps with the above correction there is no contradiction any more...

lips/lip-6.mdx
valerini: I moved the sentences about libra_root managing the validator owners up into the Workflows section.

lips/lip-6.mdx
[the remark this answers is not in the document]
valerini: I hope I rephrased correctly. I created a separate subsection on how the LibraRoot account is managed.
In case you would like to take a shot at re-phrasing it, feel free to do so! I'd also be happy to make changes upon your suggestions.

lips/lip-6.mdx
```diff
@@ -0,0 +1,140 @@
...
+* *Validator Operator* is an owner of an on-chain Libra account, it may operate none, one or several Validator Nodes: a Validator Owner may designate a Validator Operator to run a Validator Node on Validator Owner's behalf. A Validator Operator may run multiple Validator Nodes. The Validator Operator for a Validator Owner has permission to rotate the Validator Config for the Validator Owner locally, as well as copy it to the Validator Set.
+* *Reconfiguration* defines a new epoch by triggerring a new epoch event (`LibraConfig::NewEpochEvent`), it gets emitted if any part of the global on-chain configuration, such as the Validator Set, changes. If a transaction causes an epoch change, that transaction is the last transaction committed in the current epoch — any subsequent transactions in that block or future blocks from that epoch will be ignored. Once the transaction has been committed, the new configuration starts being valid immediately. E.g. the new Validator Set will start running the consensus protocol in the new epoch.
+
+<img alt="Validator Set" src={useBaseUrl('img/validator_set.svg')} />
```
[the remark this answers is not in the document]
valerini: I was trying to insert an image describing the terminology, but it is not rendering correctly indeed, will try to fix!

lips/lip-6.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+The following script relevant to validator's operations can also be run by any LibraAccount's owner:
+* [`rotate_authentication_key`](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/rotate_authentication_key.move) - to rotate the authentication key of the Validator Owner Account
+
+# Workflows
+
+This section details the workflows or common scenarios related to Validator Set management.
+## Validator Owner and Validator Operator accounts creation
+LibraRoot (through the Association Operation Services, "AOS") creates Validator Owner and Validator Operator accounts. Currently, Validator Owners are not maintaining ownership of their Libra accounts. The LibraRoot account sends Validator Owners' transaction on their behalf via AdminScripts. The Validator Owner's account address is derived from the name of the Validator Owner, the auth-key is all-zeroes vector.n The LibraRoot creates the Validator Owner's account by sending the `create_validator_account` transaction script.
```
dahliamalkhi, line 107: Once this LIP moves to Approved stage, it cannot be changed. Therefore, I don't think it is a good idea to include comments on temporary situations, like the current owners lack of key custody.

dahliamalkhi (commented):
I wonder when is the TODO going to be addressed so we can move this to "Last Call" stage?

lips/lip-6.mdx
valerini: I don't have a strong opinion here, but I see the concern of @davidiw in that this LIP is currently the only place to describe how the system is run and if we omit it from here, then this major caveat is likely to go under the radars.

What is the procedure to update a LIP? In case e.g. the design of the validator set changes? Could these lips have versions (similar to RFC, NIST standards)?

One more question: do we plan to have a LIP on how the AOS is being run? I.e. who are the approvers, how HSMs are involved, etc.?

valerini: I'll talk to David Dill tomorrow, Thu, to remove the TODO.

davidiw: I think this might be a bit too much detail though. I think we should speak on the hand of what the abstraction does here even if the implementation isn't quite there.

I don't want to block progress on the PR, so I'll approve this PR but also leave this blurb (that can be updated in a later PR or even by me!)

LibraRoot creates good...

What is required to create an account? A public key and a name (not there)

However Owners may desire LibraRoot to maintain the accounts on their behalf ...

How are account addresses derived ...

lips/lip-6.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+Validator Operators on the other hand maintain full ownership of their Libra accounts. Validator Operators should contact the AOS off-chain and communicate their authentication keys as well as the names of their organizations written in snake case (lowercase, where spaces are replaced with underscores). The authentication key is a 32 elements byte array, derived from the public verification key. The protocol currently supports Ed25519 and MultiEd25519 signature schemes. The account's address for Validator Operators is derived in a usual way.  The LibraRoot creates the Validator Owner's account by sending the `create_validator_operator_account` transaction script.
```
davidiw, line 109: yeah so my reading is we should probably have 3 paragraphs

davidiw: Hmm.. I wanted to say how account addresses are defined is an implementation detail... but that might be too hand wavy. I want to think about this more, but this okay.

i do think this means we should have 3 paragraphs, generalized account creation process and then how it differs for the two.

lips/lip-6.mdx
```diff
@@ -0,0 +1,253 @@
...
+  * Parameters:
+    * `new_key: vector<u8>`
+  * Events: none
+  * Errors:
+    * Account does not exist: `Error::make(category = Errors::NOT_PUBLISHED, reason = LibraAccount::EACCOUNT)`
+    * Key rotation capability was delegated to another account: `Error::make(category = Errors::INVALID_STATE, reason = LibraAccount::EKEY_ROTATION_CAPABILITY_ALREADY_EXTRACTED)`
+    * Malformed authentication key: `Error::make(category = Errors::INVALID_ARGUMENT, reason = LibraAccount::EMALFORMED_AUTHENTICATION_KEY)`
+
+## Formal verification guarantees on the Move Implementation
+TODO: These may change as verification work proceeds.
```
davidiw, line 237: @DavidLDill, when do you think this could be wrapped up?

davidiw (approved):

lips/lip-6.mdx
dahliamalkhi: @valerini I agree with @davidiw  's approach. we can say "owner may delegate to LibraRoot" or "LibraRoot may initially create owner account with auto delegation to LibraRoot and revoke it when Owners notify the association they have established custody of their own keys".

lips/lip-6.mdx
valerini: Done here after talking to DavidDill.

lips/lip-6.mdx
valerini: Sounds good, I tried to make it 3 paragraphs.

lips/lip-6.mdx
valerini: Sounds good, I tried to rephrase this large paragraph, hope it looks better now.

lips/lip-6.mdx
```diff
@@ -0,0 +1,139 @@
...
+* *Reconfiguration* defines a new epoch by triggerring a new epoch event (`LibraConfig::NewEpochEvent`), it gets emitted if any part of the global on-chain configuration, such as the Validator Set, changes. If a transaction causes an epoch change, that transaction is the last transaction committed in the current epoch — any subsequent transactions in that block or future blocks from that epoch will be ignored. Once the transaction has been committed, the new configuration starts being valid immediately. E.g. the new Validator Set will start running the consensus protocol in the new nepoch.
+* *LibraRoot* is an account that may add/remove members to/from the ValidatorSet, it may also send an AdminScript.
+* *AdminScript* is a transaction script that can be sent by the LibraRoot on behalf of any other account.
+
+<img alt="Validator Set" src={useBaseUrl('img/validator_set.png')} />
```
davidiw, line 33: update this, it is blocking your pr

JoelMarcey: You need to add 

```
import useBaseUrl from '@docusaurus/useBaseUrl';
``` 

towards the top of the `lip-6.mdx` file, right below the YAML metadata.

lips/lip-6.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+## Properties of the Validator Set enforced off-chain
+
+The following properties are being enforced via contractual obligations:
+* Each Validator Owner is obliged to notify LN about the decision to change the Validator Operator. LN may disallow the switch when it would violate LN's obligations, such as when a new Validator Operator is deemed untrustworthy, or the change unacceptably concentrates consensus power among too few Validator Operators or Validator Nodes' architectures.
+* If a Validator Operator OP_1 is designated to run Validator Nodes for two Validator Owners V_1 and V_2, the Validator Operator must run two distinct instances of Validator Nodes to increase fault-tolerance of the system.
+* The Validator Operator designated by the Validator Owner is responsible for running a Validator Node with the config in the Validator Set and is responsible for updating/rotating this config as needed.
+
```
davidiw, line 142: extra nl at end

davidiw:
This is sooooo awesome. At this point, I'd like to start looking at deltas on top of this instead of just revising the core content.

PR #70, opened 2020-10-02
Added anchors and labels to the tables

junkil-park:
- Added the HTML anchors (e.g., `<a name =...></a>`) and the column/row labels (e.g.A,B,C,...,1,2,3,...) for the tables so that some external documents (e.g., the Libra Framework documents) can refer to the indivisual cells of the tables.

- Changed some permission names to better reflect their associated actions
  - SetValidatorOperator --> {Set,Remove}ValidatorOperator
    - This permission is associated with not only the "set" action but also the "remove" action
  - SendWriteSetTransaction --> ProcessWriteSetTransaction
    - One needs this permission to "process" such transactions rather than "send". Anyone can "send" such transactions, but only the ones with this permission can be processed.

- Gave a more precise description about the uniqueness of the ValidatorOperator roles:
  - from "not unique" to "At most one per Validator"


dahliamalkhi (approved):
LGTM

junkil-park:
> LGTM

Thanks, @dahliamalkhi . It seems that I don't have the permission to merge. Does anyone know how I can proceed?

sblackshear (approved):

PR #75, opened 2020-10-12
Update lip-8

danprinz:
Disclaimer regarding the discovery phase
Extended scope for fund pull pre-approval
Revoke an authorization in more details

### DIFF
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -56,6 +56,7 @@ All requests between VASPs are structured as [`CommandRequestObject`s](https://l
             "biller_address": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pg9skzctpv9skzcg9kmwta",
             "funds_pre_approval_id": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_7b8404c986f53fe072301fe950d030de",
             "scope": {
+                "type": "consent",
                 "expiration_timestamp": 72322, 
                 "max_cumulative_amount": {
                     "unit": "week",
@@ -161,6 +162,14 @@ Also, the scope limits the FundPullPreApprovalObject to certain parameters of ti
     "scope": {
         "type": "consent",
         "expiration_timestamp": 72322,
+        "max_cumulative_amount": {
+            "unit": "year",
+            "value": 1,
+            "max_amount": {
+                "amount": 100000,
+                "currency": "LBR",
+            }
+        },        
         "max_transaction_amount": {
             "amount": 100,
             "currency": "LBR",
@@ -175,7 +184,7 @@ This object describes the scope of an amount
 
 | Field         | Type       | Required?    | Description   |
 |-------        |--------    |-----------   |-------------  |
-| unit          | str enum   | N            | One of: "week", "month", "year" |
+| unit          | str enum   | N            | One of: "day", "week", "month", "year" |
 | value         | int        | N            | Number of "unit"(s)   |
 | max_amount | [CurrencyObject](#currencyobject) | N | Max cumulative amount that is approved for funds pre-approval.  This is the sum across all transactions that occur in the scope of the unit value. |
 
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -176,7 +176,7 @@ This object describes the scope of an amount
 | Field         | Type       | Required?    | Description   |
 |-------        |--------    |-----------   |-------------  |
 | unit          | str enum   | N            | One of: "week", "month", "year" |
-| value         | int        | N            | "Unit" value  |
+| value         | int        | N            | Number of "unit"(s)   |
 | max_amount | [CurrencyObject](#currencyobject) | N | Max cumulative amount that is approved for funds pre-approval.  This is the sum across all transactions that occur in the scope of the unit value. |
 
 ```
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -143,7 +143,7 @@ The structure in this object can be a full pre-approval or just the fields of an
 
 ### ScopeObject
 
-In this object the initiator VASP declares its intent for the pre-approval, this can by one of two options:
+In this object the initiator VASP declares its intent for the pre-approval, this can be one of two options:
   1. Save the consumer sub-account for future transactions (save_sub_account)- this will enable the initiator VASP (merchant) to charge the sub-account in the future but will require the owner to approve the transaction. When using this option, amount limits are not required
   2. Save the consumer sub-account and get a consent for future payments (consent) - this enables the initiator VASP (merchant) to charge the sub-account without any interaction with the owner. 
 Also, the scope limits the FundPullPreApprovalObject to certain parameters of time and amount. this object can be changed by the initiator VASP if needed, but any change requires the target VASP to approve the scope change.
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -288,4 +288,4 @@ This LIP describes two independent phases of a payment - pre-approval and auth/c
 
 The first one (pre-approval) may be canceled by each side (buyer or seller). a reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
 
-The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee to get funds if requested. This has no reason to be canceled by the buyer. a request for such cancellation from the buyer should be rejected. 
+The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. a request for such cancellation from the buyer should be rejected. 
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -146,7 +146,7 @@ The structure in this object can be a full pre-approval or just the fields of an
 In this object the initiator VASP declares its intent for the pre-approval, this can be one of two options:
   1. Save the consumer sub-account for future transactions (save_sub_account)- this will enable the initiator VASP (merchant) to charge the sub-account in the future but will require the owner to approve the transaction. When using this option, amount limits are not required
   2. Save the consumer sub-account and get a consent for future payments (consent) - this enables the initiator VASP (merchant) to charge the sub-account without any interaction with the owner. 
-Also, the scope limits the FundPullPreApprovalObject to certain parameters of time and amount. this object can be changed by the initiator VASP if needed, but any change requires the target VASP to approve the scope change.
+Also, the scope limits the FundPullPreApprovalObject to certain parameters of time and amount. This object can be later mutated by the initiator VASP if needed, but any change requires the target VASP to approve the scope change.
 
 
 | Field         | Type  | Required?     | Description   |
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -151,7 +151,7 @@ Also, the scope limits the FundPullPreApprovalObject to certain parameters of ti
 
 | Field         | Type  | Required?     | Description   |
 |-------        |------ |-----------    |-------------  |
-| Type | str enum | Y | This can be either save_sub_account or consent |
+| type | str enum | Y | This can be either save_sub_account or consent |
 | expiration_timestamp | uint | Y | Unix timestamp indicating the time at which this pre-approval will expire - after which no funds pull can occur.  To expire an existing pre-approval early, this field can be updated with the current Unix timestamp. |
 | max_cumulative_amount | [ScopedCumulativeAmountObject](#scopedcumulativeamountobject) | N | Max cumulative amount that is approved for funds pre-approval.  This is the sum across all transactions that occur while utilizing this pre-approval. |
 | max_transaction_amount | [CurrencyObject](#currencyobject) | N | Max transaction amount that is approved for funds pre-approval.  This is the maximum transaction that occurs while utilizing this funds pre-approval. |
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -119,7 +119,7 @@ For a fund pre-approval request, the [command_type](https://lip.libra.org/lip-1/
 
 ### FundPullPreApprovalObject
 
-The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can be updated multiple times. Updating immutable fields with different values results in a command error, but it is acceptable to re-send the same value.
+The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can be updated multiple times. Updating immutable fields with different values results in a command error, but it is acceptable to re-send the same value. It should be noted that initial creation of the FundPullPreApprovalObject can be created or updated by the VASP on either side (buyer or seller).
 
 | Field         | Type  | Required?     | Description   |
 |-------        |------ |-----------    |-------------  |
@@ -253,7 +253,7 @@ In addition, the "_reads" field of the PaymentObject must contain the latest ver
 
 Authorization allows the placing of holds on funds with the assurance that an amount up to the held amount can be captured at a later time.  An example of this is for delayed fulfillment or pre-authorizing an expected amount to ensure that an amount can be charged after services are rendered.
 
-When an authorization happens, the sender VASP (buyer) agrees to the authorization request must lock the funds for the specified period. That is - the buyer VASP guarantees that the funds will be available if later captured.
+When an authorization happens, the sender VASP (buyer) who agrees to the authorization request must lock the funds for the specified period. That is - the buyer VASP guarantees that the funds will be available if later captured.
 
 Auth/capture is an extension of [PaymentCommand](https://lip.libra.org/lip-1/#paymentcommand-object).  The extension happens primarily within the [PaymentActionObject](https://lip.libra.org/lip-1/#paymentactionobject) and the status changes within the [PaymentActor](https://lip.libra.org/lip-1/#paymentactorobject).
 
@@ -295,6 +295,6 @@ The auth/capture flow now adds the following to the status enum of [PaymentActor
 
 This LIP describes two independent phases of a payment - pre-approval and auth/capture.
 
-The first one (pre-approval) may be canceled by each side (buyer or seller). a reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
+The first one (pre-approval) may be canceled by either side (buyer or seller). A reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
 
-The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. a request for such cancellation from the buyer should be rejected. 
+The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. A request for such cancellation from the buyer should be rejected. 
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -22,7 +22,7 @@ Version 0 of the Off-Chain Protocol is described in [LIP 1](https://lip.libra.or
 ---
 
 This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
-The process describes below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
+The process described below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
 For the discovery phase, there are currently two leading methods:
   1. The buyer VASP triggers a process with pre-knowledge of the merchant subaddress. e.g., QR/deep links - where the consumers while acting in their wallet provides their VASP the merchant's relevant details (embedded in the QR/link in the checkout page)
   2. The merchant gets the user identifier (like Pay ID) and query the buyer side for a new subaddress for this process
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -3,7 +3,7 @@ lip: 8
 title: Off-chain API Version 1 Extensions
 authors: Kevin Hurley (@kphfb)
 status: Draft
-type: Informational
+type: Standard
 created: 09/18/2020
 ---
 
@@ -23,10 +23,7 @@ Version 0 of the Off-Chain Protocol is described in [LIP 1](https://lip.libra.or
 
 This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
 The process described below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
-For the discovery phase, there are currently two leading methods:
-  1. The buyer VASP triggers a process with pre-knowledge of the merchant subaddress. e.g., QR/deep links - where the consumers while acting in their wallet provides their VASP the merchant's relevant details (embedded in the QR/link in the checkout page)
-  2. The merchant gets the user identifier (like Pay ID) and query the buyer side for a new subaddress for this process
-For now, we will assume that one of these methods took place and the merchant have the buyer subaddress at hand.
+For now, we will assume that such a discovery process took place and the merchant have the buyer subaddress at hand.
   
 ---
 # Specification
@@ -226,7 +223,7 @@ Valid values are:
 * `rejected` - User/VASP did not approve the pre-approval request.
 * `closed` - Approval has been closed by the user/VASP and can no longer be used.
 
-**Valid Status Transitions**. Either party in the pre-approval agreement can mutate the status. The status always initially begins as `pending` at which time a user must agree to the pre-approval request.  Once the user has reviewed the request, the biller VASP will update the pre-approval status to `valid` (if the user agreed) or `rejected` (if the user rejected the pre-approval).
+**Valid Status Transitions**. Either party in the pre-approval agreement can mutate the status. The status always initially begins as `pending` at which time a user must agree to the pre-approval request.  Once the user has reviewed the request, the sender VASP (of the buyer) will update the pre-approval status to `valid` (if the user agreed) or `rejected` (if the user rejected the pre-approval).
 
 At any point, the user can withdraw permission at which point the status will be updated to `closed`.
 
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -154,7 +154,7 @@ Also, the scope limits the FundPullPreApprovalObject to certain parameters of ti
 | type | str enum | Y | This can be either save_sub_account or consent |
 | expiration_timestamp | uint | Y | Unix timestamp indicating the time at which this pre-approval will expire - after which no funds pull can occur.  To expire an existing pre-approval early, this field can be updated with the current Unix timestamp. |
 | max_cumulative_amount | [ScopedCumulativeAmountObject](#scopedcumulativeamountobject) | N | Max cumulative amount that is approved for funds pre-approval.  This is the sum across all transactions that occur while utilizing this pre-approval. |
-| max_transaction_amount | [CurrencyObject](#currencyobject) | N | Max transaction amount that is approved for funds pre-approval.  This is the maximum transaction that occurs while utilizing this funds pre-approval. |
+| max_transaction_amount | [CurrencyObject](#currencyobject) | N | Max transaction amount that is approved for funds pre-approval.  This is the maximum amount that may occur in a single transaction while utilizing this funds pre-approval. |
 
 ```
 {
--- a/lips/lip-8.md
+++ b/lips/lip-8.md
@@ -18,13 +18,24 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
 Version 0 of the Off-Chain Protocol is described in [LIP 1](https://lip.libra.org/lip-1/).  Version 1 as described here is an extension of the Off-Chain Protocol and adds features to support more advanced functionality - particularly targeted to support merchant use-cases.  This is inclusive of pull payments, recurring payments, and auth/capture flows.
 
 ---
+# Disclaimer
+---
+
+This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
+The process describes below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
+For the discovery phase, there are currently two leading methods:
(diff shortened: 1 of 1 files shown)


JoelMarcey:
@danprinz Looks like a rebase may be needed.

@bmwill Is there something we need to do with `bors` here? I don't usually see that warning in PRs to this repo, I don't think.

bmwill:
@JoelMarcey Did we ever  resolve if we wanted bors on this repo or not? If not I probably need to just disable bors since IIRC we aren't currently relying on bors for this repo.

JoelMarcey:
@bmwill I think I am ok removing `bors` on this repo for now and revisit later. We have the passive checks to make sure the website builds and we have only `CODEOWNERS` being able to merge, which helps too.

lips/lip-8.md
```diff
@@ -17,14 +17,25 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
...
+# Disclaimer
+---
+
+This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
+The process describes below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
+For the discovery phase, there are currently two leading methods:
+  1. The buyer VASP triggers a process with pre-knowledge of the merchant subaddress. e.g., QR/deep links - where the consumers while acting in their wallet provides their VASP the merchant's relevant details (embedded in the QR/link in the checkout page)
+  2. The merchant gets the user identifier (like Pay ID) and query the buyer side for a new subaddress for this process
+For now, we will assume that one of these methods took place and the merchant have the buyer subaddress at hand.
+
```
kphfb, line 30: I think we should leave this portion out.  It's outside the scope of this LIP and I think it's enough to mention that it isn't in the scope of this LIP without speculating how it may be done

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
 }
 ```
 
+### ScopeObject
+
+In this object the initiator VASP declares its intent for the pre-approval, this can by one of two options:
```
kphfb, line 146:
```suggestion
In this object the initiator VASP declares its intent for the pre-approval, this can be one of two options:
```

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+  1. Save the consumer sub-account for future transactions (save_sub_account)- this will enable the initiator VASP (merchant) to charge the sub-account in the future but will require the owner to approve the transaction. When using this option, amount limits are not required
+  2. Save the consumer sub-account and get a consent for future payments (consent) - this enables the initiator VASP (merchant) to charge the sub-account without any interaction with the owner. 
+Also, the scope limits the FundPullPreApprovalObject to certain parameters of time and amount. this object can be changed by the initiator VASP if needed, but any change requires the target VASP to approve the scope change.
```
kphfb, line 149:
```suggestion
Also, the scope limits the FundPullPreApprovalObject to certain parameters of time and amount. This object can be later mutated by the initiator VASP if needed, but any change requires the target VASP to approve the scope change.
```

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+
+
+| Field         | Type  | Required?     | Description   |
+|-------        |------ |-----------    |-------------  |
+| Type | str enum | Y | This can be either save_sub_account or consent |
```
kphfb, line 154:
```suggestion
| type | str enum | Y | This can be either save_sub_account or consent |
```

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+| expiration_timestamp | uint | Y | Unix timestamp indicating the time at which this pre-approval will expire - after which no funds pull can occur.  To expire an existing pre-approval early, this field can be updated with the current Unix timestamp. |
+| max_cumulative_amount | [ScopedCumulativeAmountObject](#scopedcumulativeamountobject) | N | Max cumulative amount that is approved for funds pre-approval.  This is the sum across all transactions that occur while utilizing this pre-approval. |
+| max_transaction_amount | [CurrencyObject](#currencyobject) | N | Max transaction amount that is approved for funds pre-approval.  This is the maximum transaction that occurs while utilizing this funds pre-approval. |
```
kphfb, line 157:
```suggestion
| max_transaction_amount | [CurrencyObject](#currencyobject) | N | Max transaction amount that is approved for funds pre-approval.  This is the maximum amount that may occur in a single transaction while utilizing this funds pre-approval. |
```

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+
+```
+{
+    "scope": {
```
kphfb, line 161: Let's add the max_cumulative_amount to this also

lips/lip-8.md
```diff
@@ -43,11 +54,17 @@ All requests between VASPs are structured as [`CommandRequestObject`s](https://l
         "fund_pull_pre_approval": {
             "address": "lbr1pgfpyysjzgfpyysjzgfpyysjzgf3xycnzvf3xycsm957ne",
             "biller_address": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pg9skzctpv9skzcg9kmwta",
-            "funds_pre_approval_id": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_7b8404c986f53fe072301fe950d030de"
-            "expiration_timestamp": 72322, 
-            "max_cumulative_amount": {
-                "amount": 1000,
-                "currency": "LBR"
+            "funds_pre_approval_id": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_7b8404c986f53fe072301fe950d030de",
+            "scope": {
```
kphfb, line 58: This doesn't have "type" in it

lips/lip-8.md
kphfb, line 154: I'm not sure I understand why we would want "type".  save_sub_account seems to not be overly useful, but perhaps I'm missing something.  Isn't it implied that if I wanted to bill someone again, I could try their same subaddress and they would have the option to approve it?  Even if I didn't get this approval ahead of time, I could still trigger the transaction, couldn't I?

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+}
+```
+
+### ScopedCumulativeAmountObject
+
+This object describes the scope of an amount
+
+| Field         | Type       | Required?    | Description   |
+|-------        |--------    |-----------   |-------------  |
+| unit          | str enum   | N            | One of: "week", "month", "year" |
```
kphfb, line 178: Should we add "day"?  And what about something for an all-time limit - or should a high number of years suffice for that?

lips/lip-8.md
```diff
@@ -125,12 +141,65 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+| value         | int        | N            | "Unit" value  |
```
kphfb, line 179:
```suggestion
| value         | int        | N            | Number of "unit"(s)   |
```

lips/lip-8.md
```diff
@@ -205,8 +276,16 @@ The [PaymentActionObject](https://lip.libra.org/lip-1/#paymentactionobject) now
...
-**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if one side wishes to cancel the auth).
+**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if a valid cancel request was sent).
+
+## Cancellation
+
+This LIP describes two independent phases of a payment - pre-approval and auth/capture.
+
+The first one (pre-approval) may be canceled by each side (buyer or seller). a reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
+
+The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee to get funds if requested. This has no reason to be canceled by the buyer. a request for such cancellation from the buyer should be rejected.
```
kphfb, line 291:
```suggestion
The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. a request for such cancellation from the buyer should be rejected. 
```

kphfb (changes requested):

lips/lip-8.md
danprinz: for "days" - you're right. I will add it. 
for "all-time" - I would go with the latter to keep the structure the same for all options (if we set a unit value of "all-time" then we don't need a "value" field which makes parsing uglier). not crucial, just a matter of taste.

lips/lip-8.md
danprinz: Agreed. I will remove it.

lips/lip-8.md
danprinz: you are right that "consent" covers the basic option of "save_sub_account". 
but we thought it would be nice to have the user approve even the storage of their details (sub-account). 
consider the scenario of users who wants to add their Libra wallet as a payment method (to make subsequent payments easier), but still want to authorize every payment. something like a push notification to their mobile app. 
the merchant VASP should store their sub-address but without a "real" consent of withdrawing funds. 
one may argue that the merchant could save these details without prior approval of the user, but I think that users will want to authorize even this. 
the wallet VASPs on the other hand will pop a notification to the user only if the merchant could present such an approval.    

does it make sense?

lips/lip-8.md
```diff
@@ -17,14 +17,25 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
 
 Version 0 of the Off-Chain Protocol is described in [LIP 1](https://lip.libra.org/lip-1/).  Version 1 as described here is an extension of the Off-Chain Protocol and adds features to support more advanced functionality - particularly targeted to support merchant use-cases.  This is inclusive of pull payments, recurring payments, and auth/capture flows.
 
+---
+# Disclaimer
+---
+
+This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
+The process described below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
+For the discovery phase, there are currently two leading methods:
```
dahliamalkhi, line 26: I am not sure we want to include any statements here on discovery methods, since this is an evolving area and there is currently no method(s) that Libra has standardized. I would rephrase and say that the industry has mechanisms such as [...] and others.

lips/lip-8.md
```diff
@@ -17,14 +17,25 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
...
+  1. The buyer VASP triggers a process with pre-knowledge of the merchant subaddress. e.g., QR/deep links - where the consumers while acting in their wallet provides their VASP the merchant's relevant details (embedded in the QR/link in the checkout page)
+  2. The merchant gets the user identifier (like Pay ID) and query the buyer side for a new subaddress for this process
+For now, we will assume that one of these methods took place and the merchant have the buyer subaddress at hand.
+  
 ---
 # Specification
```
dahliamalkhi, line 32: I am missing a high level description of the proposed approach, before we jump into the Specification.

dahliamalkhi (changes requested):

lips/lip-8.md
```diff
@@ -17,14 +17,25 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
 
 Version 0 of the Off-Chain Protocol is described in [LIP 1](https://lip.libra.org/lip-1/).  Version 1 as described here is an extension of the Off-Chain Protocol and adds features to support more advanced functionality - particularly targeted to support merchant use-cases.  This is inclusive of pull payments, recurring payments, and auth/capture flows.
 
+---
+# Disclaimer
```
dahliamalkhi, line 21: Now that Libra 1.0.0 is released, this LIP should not be Informational but Standard.

lips/lip-8.md
```diff
@@ -17,14 +17,25 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
...
+---
+
+This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
+The process described below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
+For the discovery phase, there are currently two leading methods:
+  1. The buyer VASP triggers a process with pre-knowledge of the merchant subaddress. e.g., QR/deep links - where the consumers while acting in their wallet provides their VASP the merchant's relevant details (embedded in the QR/link in the checkout page)
+  2. The merchant gets the user identifier (like Pay ID) and query the buyer side for a new subaddress for this process
+For now, we will assume that one of these methods took place and the merchant have the buyer subaddress at hand.
+
```
kphfb, line 30: Yeah, I think we should just leave this out since it's speculation

lips/lip-8.md
kphfb: Hmmm, I'm still not convinced that this is something that merchants/users will want.  I could be wrong though.  But it feels like there is no real reason for anyone to follow this for save_sub_account.  I can just put a checkbox on my page to "save" in the same way and it functionally operates exactly the same as this without needing any off-chain or user action in their wallet app.  It just feels like we are adding a feature that isn't actually adding anything practically, but does add some complexity in implementation

lips/lip-8.md
```diff
@@ -148,7 +226,7 @@ Valid values are:
 * `rejected` - User/VASP did not approve the pre-approval request.
 * `closed` - Approval has been closed by the user/VASP and can no longer be used.
 
-**Valid Status Transitions**. Either party in the pre-approval agreement can mutate the status. The status always initially begins as `pending` at which time a user must agree to the pre-approval request.  Once the user has reviewed the request, the billee VASP will update the pre-approval status to `valid` (if the user agreed) or `rejected` (if the user rejected the pre-approval).
+**Valid Status Transitions**. Either party in the pre-approval agreement can mutate the status. The status always initially begins as `pending` at which time a user must agree to the pre-approval request.  Once the user has reviewed the request, the biller VASP will update the pre-approval status to `valid` (if the user agreed) or `rejected` (if the user rejected the pre-approval).
```
kphfb, line 229: This would still be billee.  Or maybe we should user "payer" and "payee" instead.  In which case, this would be the payer VASP.  The side that has the user on it.  Actually, let's just call it the sender VASP (buyer) maybe

lips/lip-8.md
```diff
@@ -175,16 +253,18 @@ In addition, the "_reads" field of the PaymentObject must contain the latest ver
 
 Authorization allows the placing of holds on funds with the assurance that an amount up to the held amount can be captured at a later time.  An example of this is for delayed fulfillment or pre-authorizing an expected amount to ensure that an amount can be charged after services are rendered.
 
-When an authorization happens, the VASP agreeing to the authorization must lock the funds for the specified amount of time - the VASP is agreeing to a guarantee that the funds will be available if later captured.
+When an authorization happens, the sender VASP (buyer) agrees to the authorization request must lock the funds for the specified period. That is - the buyer VASP guarantees that the funds will be available if later captured.
```
kphfb, line 256:
```suggestion
When an authorization happens, the sender VASP (buyer) who agrees to the authorization request must lock the funds for the specified period. That is - the buyer VASP guarantees that the funds will be available if later captured.
```

lips/lip-8.md
```diff
@@ -205,8 +285,16 @@ The [PaymentActionObject](https://lip.libra.org/lip-1/#paymentactionobject) now
...
 `abort` may still be used to cancel the authorization early.  Once a capture action occurs, the status of the payment will now be updated to `ready_for_settlement`.
 
-**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if one side wishes to cancel the auth).
+**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if a valid cancel request was sent).
+
+## Cancellation
+
+This LIP describes two independent phases of a payment - pre-approval and auth/capture.
+
+The first one (pre-approval) may be canceled by each side (buyer or seller). a reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel).
```
kphfb, line 298:
```suggestion
The first one (pre-approval) may be canceled by either side (buyer or seller). A reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
```

lips/lip-8.md
```diff
@@ -205,8 +285,16 @@ The [PaymentActionObject](https://lip.libra.org/lip-1/#paymentactionobject) now
...
-**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if one side wishes to cancel the auth).
+**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if a valid cancel request was sent).
+
+## Cancellation
+
+This LIP describes two independent phases of a payment - pre-approval and auth/capture.
+
+The first one (pre-approval) may be canceled by each side (buyer or seller). a reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
+
+The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. a request for such cancellation from the buyer should be rejected.
```
kphfb, line 300:
```suggestion
The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. A request for such cancellation from the buyer should be rejected. 
```

lips/lip-8.md
```diff
@@ -101,16 +119,15 @@ For a fund pre-approval request, the [command_type](https://lip.libra.org/lip-1/
 
 ### FundPullPreApprovalObject
 
-The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can by updated multiple times. Updating immutable fields with a different value results in a command error, but it is acceptable to re-send the same value.
+The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can be updated multiple times. Updating immutable fields with different values results in a command error, but it is acceptable to re-send the same value.
```
kphfb, line 122:
```suggestion
The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can be updated multiple times. Updating immutable fields with different values results in a command error, but it is acceptable to re-send the same value. It should be noted that initial creation of the FundPullPreApprovalObject can be created or updated by the VASP on either side (buyer or seller).
```

kphfb (changes requested):

kphfb:
Think with the one change I suggest, it hopefully addresses https://github.com/libra/lip/issues/71 as well.  If so, we can close that one once this is updated

lips/lip-8.md
```diff
@@ -17,14 +17,22 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
 
 Version 0 of the Off-Chain Protocol is described in [LIP 1](https://lip.libra.org/lip-1/).  Version 1 as described here is an extension of the Off-Chain Protocol and adds features to support more advanced functionality - particularly targeted to support merchant use-cases.  This is inclusive of pull payments, recurring payments, and auth/capture flows.
 
+---
+# Disclaimer
+---
+
+This LIP does not contain the initial phase of a sub-account discovery that is required to start negotiating merchant scenarios.
+The process described below starts at the phase that both sides (particularly the biller side) have the relevant subaddresses. 
+For now, we will assume that such a discovery process took place and the merchant have the buyer subaddress at hand.
```
kphfb, line 26:
```suggestion
For now, we will assume that such a discovery process took place and the merchant has the buyer subaddress at hand.
```

kphfb:
I think I'm mostly good with this PR other than I'm still not convinced about "type" and want to chat about that first

lips/lip-8.md
kphfb: @dahliamalkhi what type of thing would you like to see here as compared to what is there in the abstract section?

lips/lip-8.md
```diff
@@ -205,8 +282,16 @@ The [PaymentActionObject](https://lip.libra.org/lip-1/#paymentactionobject) now
 
 The auth/capture flow now adds the following to the status enum of [PaymentActor](https://lip.libra.org/lip-1/#paymentactorobject):
 
-* `authorized` - Payment amount is authorized but not yet captured.
+* `authorized` - The payment amount is authorized but not yet captured.
 
 `abort` may still be used to cancel the authorization early.  Once a capture action occurs, the status of the payment will now be updated to `ready_for_settlement`.
 
-**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if one side wishes to cancel the auth).
+**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if a valid cancel request was sent).
```
LuZhang-Lou, line 289: after reading the whole auth/capture section, I still do not understand how the valid status transition. Can we have a detailed diagraph like we are making for lip-1?

lips/lip-8.md
```diff
@@ -125,12 +139,73 @@ The structure in this object can be a full pre-approval or just the fields of an
 }
 ```
 
+### ScopeObject
+
+In this object the initiator VASP declares its intent for the pre-approval, this can be one of two options:
+  1. Save the consumer sub-account for future transactions (save_sub_account)- this will enable the initiator VASP (merchant) to charge the sub-account in the future but will require the owner to approve the transaction. When using this option, amount limits are not required
+  2. Save the consumer sub-account and get a consent for future payments (consent) - this enables the initiator VASP (merchant) to charge the sub-account without any interaction with the owner. 
+Also, the scope limits the FundPullPreApprovalObject to certain parameters of time and amount. This object can be later mutated by the initiator VASP if needed, but any change requires the target VASP to approve the scope change.
```
xli, line 147: what is "sub_account" means?
use name "type" is confusing here, as type indicates the object type, which is protocol level,
but it looks like a business agreement about how the next payment object can be processed.

lips/lip-8.md
```diff
@@ -125,12 +139,73 @@ The structure in this object can be a full pre-approval or just the fields of an
...
+```
+
+### ScopedCumulativeAmountObject
+
+This object describes the scope of an amount
+
+| Field         | Type       | Required?    | Description   |
+|-------        |--------    |-----------   |-------------  |
+| unit          | str enum   | N            | One of: "day", "week", "month", "year" |
+| value         | int        | N            | Number of "unit"(s)   |
```
xli, line 185: what is difference between above 2 fields with "expiration_timestamp"? they are also confusing because different application may have different interpretation about what means of start of the day/week/month/year, hence not accurate.

danprinz: These fields are for limiting (scoping) the fund pulls from the consumer VASP. It can't be expressed using an expiration time, as it might be "not more than $100 a week". 
Regarding the interpretation of week start, duration, etc. - we think of it as a sliding time window, but one may think it's a calendar timeframe.
I will add a more detailed description of it.

lips/lip-8.md
```diff
@@ -101,16 +116,15 @@ For a fund pre-approval request, the [command_type](https://lip.libra.org/lip-1/
...
+| biller_address | str | Required for creation | Address of account from which billing will happen. Addresses may be single-use or valid for a limited time, and therefore VASPs should not rely on them remaining stable across time or different VASP addresses. The addresses are encoded using bech32. The bech32 address encodes both the address of the VASP as well as the specific user's subaddress. They should be no longer than 80 characters. Mandatory and immutable. For Libra addresses, refer to the "account identifier" section in LIP-5 for format. |
 | funds_pre_approval_id | str | Y | Unique reference ID of this pre-approval on the pre-approval initiator VASP (the VASP which originally created this pre-approval object). This value should be unique, and formatted as “<creator_vasp_onchain_address_bech32>_<unique_id>”.  For example, ”lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_7b8404c986f53fe072301fe950d030de“. Note that this should be the VASP address and thus have a subaddress portion of 0. This field is mandatory on pre-approval creation and immutable after that.  Updates to an existing pre-approval must also include the previously created pre-approval ID. |
```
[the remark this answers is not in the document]
danprinz: It won't be very clear once one will try to use it in a subsequent payment (a reference to the user consent). 
IMHO this name is more descriptive.

lips/lip-8.md
danprinz: "sub_account" is the consumer identifier within (behind) the VASP. 
It might be a good idea to replace it with another term that indicates the wallet user.

lips/lip-8.md
davidiw: I think the normal approach would be to describe a general strawman flow, then explicitly state what is assumed (e.g., that there is a user -> merchant identity exchange process). As I noted in another PR on DIP-8, there's so much boiler plate before getting into what this DIP does, that it ... isn't particularly accessible. You calling this out, means you see it too :).

lips/lip-8.md
```diff
@@ -17,14 +17,22 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
...
+Establishes a relationship between the sender and recipient where the recipient can now pull funds from the sender without the sender approving each transaction.  This allows the recipient to bill the sender without the sender approving each payment.  This relationship exists between a subaddress on the biller side and a subaddress on the sender's side.  After this request is POSTed, the target VASP can use out-of-band methods to determine if this request should be granted.  If the target VASP chooses to allow the relationship to be established, the biller can create a payment object and POST to the billed party’s VASP to request funds.  The “funds_pull_approval_id” object must then match the ID established by this request.
```
davidiw, line 35: Here's another assumption, an out-of-band process, I think having a strawman definition of this protocol broken down into bullets would be more accessible. We could even allude to one such method for doing the exchanges.

lips/lip-8.md
```diff
@@ -43,11 +51,18 @@ All requests between VASPs are structured as [`CommandRequestObject`s](https://l
         "fund_pull_pre_approval": {
             "address": "lbr1pgfpyysjzgfpyysjzgfpyysjzgf3xycnzvf3xycsm957ne",
             "biller_address": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pg9skzctpv9skzcg9kmwta",
-            "funds_pre_approval_id": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_7b8404c986f53fe072301fe950d030de"
-            "expiration_timestamp": 72322, 
-            "max_cumulative_amount": {
-                "amount": 1000,
-                "currency": "LBR"
+            "funds_pre_approval_id": "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_7b8404c986f53fe072301fe950d030de",
+            "scope": {
```
davidiw, line 55: what's the value of having another data structure? could we amend the PR description or is there a tracking doc for this?

lips/lip-8.md
```diff
@@ -101,16 +116,15 @@ For a fund pre-approval request, the [command_type](https://lip.libra.org/lip-1/
 
 ### FundPullPreApprovalObject
 
-The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can by updated multiple times. Updating immutable fields with a different value results in a command error, but it is acceptable to re-send the same value.
+The structure in this object can be a full pre-approval or just the fields of an existing pre-approval object that need to be changed. Some fields are immutable after they are defined once (see below). Others can be updated multiple times. Updating immutable fields with different values results in a command error, but it is acceptable to re-send the same value. It should be noted that initial creation of the FundPullPreApprovalObject can be created or updated by the VASP on either side (buyer or seller).
```
davidiw, line 119: This seems kinda weird to me, it feels like it would make more sense for this to be something like a client submits a pre-approval to the merchant's VASP. There shouldn't be any mutable fields either it is agreeable or not. If it is not, return back some error that the service can then address. If the services should support follow up pre-approvals, those should be separate interactions not building on the same "object". I should have asked this on Kevin's PR, but I'll ask it on yours instead.

lips/lip-8.md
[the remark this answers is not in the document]
davidiw: I think the key thing here is .... how do we distinctly distinguish different objects... a reference_id is supposed to be common to all objects -- if there is a shared object to be created or maintained. Also it might behove us to pivot to DiemId rather than further embracing the jankiness of subaddresses, they are really dizzying.

lips/lip-8.md
davidiw: ahh, interesting insight, we should determine a uniform fashion to call these out. I am concerned that subaccount doesn't really refer to a client's account and I've heard a lot of confusion across subaddress and child vasps :/

lips/lip-8.md
davidiw: unit - reset_period?
value - amount_per_period

also I would probably move max_amount and currency to the same level since currency refers to both...

lips/lip-8.md
```diff
@@ -205,8 +282,16 @@ The [PaymentActionObject](https://lip.libra.org/lip-1/#paymentactionobject) now
...
-**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if one side wishes to cancel the auth).
+**Valid Status Transitions**. `authorized` is now a valid initial value and may be followed by `ready_for_settlement` (upon a successful capture) or `abort` (if a valid cancel request was sent).
+
+## Cancellation
+
+This LIP describes two independent phases of a payment - pre-approval and auth/capture.
+
+The first one (pre-approval) may be canceled by either side (buyer or seller). A reasonable scenario is when a consumer wishes to cancel a subscription (buyer cancel) or asks a merchant app to remove the user wallet from the list of payment methods (seller cancel). 
+
+The second (authorization) could be canceled only by the biller (merchant) as it holds a guarantee of funds availability if requested. This has no reason to be canceled by the buyer. A request for such cancellation from the buyer should be rejected.
```
davidiw, line 297: shouldn't there be a follow up command to cancel?

I think having a clean story of how these pieces fit together would be immensely valuable for this DIP. As a super noob to the space, I'm struggling a bit to see how all these fit together.

davidiw (commented):
Really excited to see the enthusiasm in the space! And it makes it painfully difficult how hard it can be to work effectively in github for doc related reviews :P

I think the biggest contribution here is to enable the notion of periodic resettable limits within a single approval --- that makes some sense. I'm not aware of whether or not this is a feature that existing payment providers offer, so would we potentially be increasing the barrier to entry? With that being said, I do like the concept and am wondering if it should be a different command or a follow up.

davidiw:
feel free to re-open but this is pretty stale now

PR #79, opened 2020-10-21
Image for LIP-6 on validator set management

valerini:
Pre-commit image to enable rendering from PR #66.


davidiw (approved):
I admit, i don't know the best place to put this... I'm not really a big fan of dumping everything to a single folder as it makes it harder to figure out dependencies and then you keep images around forever, but I guess since lips are forever :D

#movefast

PR #80, opened 2020-10-28
LIP-9 - private metadata

danprinz:


dahliamalkhi (commented):
My sense is that this proposal is somewhat preliminary, and needs to be discussed with legal/compliance and the Merchants WG to proceed further. 

Additionally, once wee are ready to push this forward, this should be a Standard LIP (not informational). It possibly may need to be (re)assigned a new LIP number, to be discussed with the Lead Maintainer.

lips/lip-9.md
```diff
@@ -0,0 +1,95 @@
...
+authors: Daniel Prinz
+status: Draft
+type: Informational
+created: 09/10/2020
+---
+# Summary
+This LIP describes the conceptual model and implementation of a private, on-chain description attribute, that could be added to every P2P transaction. An example of that would be short and descriptive information about the transaction, like what are the purchased items that are being paid in a transaction. 
+
+# Abstract / Overview
+The Libra blockchain can carry informational metadata with every transaction made. This data could vary and usually used for validating both sides of the transaction could "understand" its mean. While this metadata deals with the functioning of the sender and receiver VASPs, and/or with the FATF travel rule regulatory constraints, the transaction info itself remains opaque to the end-users. This property is missing from the current payment experience of any traditional network.
```
kphfb, line 13: This paragraph is unclear to me.  I'm not entirely sure what it's trying to convey

lips/lip-9.md
```diff
@@ -0,0 +1,95 @@
...
+This LIP goal is to propose a standard way of describing any transaction while keeping the following principals:
```
kphfb, line 14:
```suggestion
This LIP's goal is to propose a standard way of describing any transaction while keeping the following principles:
```

lips/lip-9.md
```diff
@@ -0,0 +1,95 @@
...
+
+encrypted_product_description = client_aes.encrypt(signed_product.serialize())
+
+# Submit the payment transaction with *encrypted_product_description* as metadata variant
+```
+
+## End-user private key security
+As with every secret private key, usual end users tend to be afraid of having such solely. In most cases, the end-user chooses to store it in a secure location and retrieve it with a more conventual method of identification such as a username/password or biometric. this feature doesn't deny such service to securely store an end user keys. it is expected to be a service supplied by the end-user wallet. it is distinguished from the wallet blockchain account private key, as it is a different one for each customer and not only one for the custodial account as a whole. this way the end-users would be able to ask their wallet to get this secret key and import it to a 3rd party software for analysis purposes. In case of end-users chooses to use their private keys, on their own risk of losing it, it should be supported, and the consequence of such a loss shouldn't be significant in terms of funds locked in an account or being lost.
```
kphfb, line 70:
```suggestion
As with every secret private key, usual end users tend to be afraid of having such solely. In most cases, the end-user chooses to store it in a secure location and retrieve it with a more conventual method of identification such as a username/password or biometric. This feature doesn't deny such service to securely store an end user keys. It is expected to be a service supplied by the end-user wallet. It is distinguished from the wallet blockchain account private key, as it is a different one for each customer and not only one for the custodial account as a whole. this way the end-users would be able to ask their wallet to get this secret key and import it to a 3rd party software for analysis purposes. In case of end-users chooses to use their private keys, on their own risk of losing it, it should be supported, and the consequence of such a loss shouldn't be significant in terms of funds locked in an account or being lost.  
```

lips/lip-9.md
```diff
@@ -0,0 +1,95 @@
+---
```
kphfb, line 1: General comments:
1. It's unclear to me why this is desirable to have on chain.  If my key were leaked, now all of my purchase information is parsable.
2. If we wanted to have this rich metadata, why wouldn't we prefer to do it over off-chain? It keeps it private and allows for richer data than what we would want to ever store on chain (for example, maybe an image)
3. I'm uncertain if users will ever really want to proactively share their purchase information with a 3rd-party.  I wonder if we should wait until we see a real use case for this?
4. One of the things you mention is portability of transaction information.  I agree that there is value there.  But not sold on this being the way to do it.  This feels more dangerous.  Why not use off-chain for dumping data and importing data?  Francois has been thinking about this as well, let me ping him to chime in here
5. For resolving disputes, wouldn't the VASP generally be able to do this?  So if you exchanged this data via off-chain, the VASPs would have the data already and could do dispute resolution
6. If we did want this to be on-chain, why include the raw data instead of just signature of the hash of the data?  That way, others couldn't reconstruct it
7. I worry that this could easily cause confusion.  Some company says "share your private key with us for $5 so we can analyze your payments" and now I accidentally share my wallet private key.  I worry that the potential risks are higher than any benefits from it

kphfb (changes requested):
Added some comments

lips/lip-9.md
```diff
@@ -0,0 +1,95 @@
...
+type: Informational
+created: 09/10/2020
+---
+# Summary
+This LIP describes the conceptual model and implementation of a private, on-chain description attribute, that could be added to every P2P transaction. An example of that would be short and descriptive information about the transaction, like what are the purchased items that are being paid in a transaction. 
+
+# Abstract / Overview
+The Libra blockchain can carry informational metadata with every transaction made. This data could vary and usually used for validating both sides of the transaction could "understand" its mean. While this metadata deals with the functioning of the sender and receiver VASPs, and/or with the FATF travel rule regulatory constraints, the transaction info itself remains opaque to the end-users. This property is missing from the current payment experience of any traditional network.
+This LIP goal is to propose a standard way of describing any transaction while keeping the following principals:
+1. Indisputable data
```
kchalkias, line 15: This property can also be satisfied by submitting a salted hash of the contents Vs the content itself (even if encrypted)

lips/lip-9.md
```diff
@@ -0,0 +1,95 @@
...
+2. Privacy of the end-user (sender)
+3. Ownership of the data by the end-user 
+
+
+# Terminology
+* *End users/Customers* are custodial or non-custodial wallet users which wish to transfer funds from their account to the beneficiary account (Receiver/Merchant)
+* *End-user wallet* is an on-chain VASP who provides custodian services to the end-users. This entity is considered optional and this role could be played by the end-users in a non-custodial wallet scenarios
+* *Receiver/Merchant VASP* is an entity that provides the same VASP services to the other side of the transaction, such as a regular receiver or merchant
+* *Receiver/Merchant* is the beneficiary of the transaction
+* *Private/Public keys* are keys used to sign and verify as well as encrypt and decrypt data by both sides. a private key is usually used to sign and encrypt while a public key is usually used for verification
```
kchalkias, line 25: Encrypting using classical public key cryptography and storing this data in a ledger (that will stay accessible for ever) can cause many issues in the future (from information disclosure to reputation).

Assume there is a cryptanalytic attack or some regulations in the future that demand a minimum key size or even post-quantum security. Unlike signatures where we can rotate keys in theory, publicly available encrypted data cannot be updated. Even if this was tolerable today (I doubt it is in practice), we don't know the situation in 10-20 years from now, so we have to be very careful with f we ever allow encryption on chain for potentially sensitive data.

lips/lip-9.md
danprinz: It meant to explain why do we need such details on-chain. 
currently, transaction metadata carrying only functional data (like travel rule reference, subaddresses, etc), but I think it should support an extra level of private data.
In legacy payment systems, the issuer usually can not present the user with great details about a specific transaction, rather some generic idea. Think of a credit card statement, in which you can see that you paid to Amazon, but you can not tell what you've bought. For that, you need to login into your account at Amazon and look for these details. What if each transaction of your wallet was carrying extra data that describes the payment in the very same way (and details) as you saw it when authorizing the payment? it would give you as a user a lot more power and control over your account. Also, it opens a new section of fin-apps to analyze your purchases on your behalf and make optimization suggestions for your benefit.

lips/lip-9.md
danprinz: Thanks for your feedback.
Generally speaking, the concept of storing such data on-chain and not off-chain is for giving the users more power and control over their data. When storing it off-chain, the user actually relies on the wallet/merchant VASP(s) to give the data when they asked to do it. This data belongs to the user, not the VASP. 
Under this assumption, let me address your points:  

1. That's right. I believe that most wallets and vendors will store your private key for you and let you identify yourself using conventional authentication methods. But when you will want to drop from your current wallet vendor and move to another one, it should be a fairly straightforward process to export your personal PK and import it to the new vendor.
2. Fair point. I want it to be on-chain for the above reasons. Going off-chain will let us store richer data.
3. There are already a large collection of 3rd party apps that scans your bank account and cc statements to give you insights and suggestions for your financial behavior. I think it would be very useful for such use cases. 
4. Making it off-chain is possible. in such a case we will have to define a standard way of export/import. The main issue will be how we can enforce these rules on the VASPs. 
5. Sure. Although the only thing that is signed is the reference_id of the off-chain protocol. If we want to move it to off-chain I think we should add one or more steps to wrap the whole payment with undisputable data from both sides. 
6. Including only hashes would still leave the actual data (and power) in the VASP(s) hands. I don't see any difference between such a hash and the currently suggested metadata (reference_id)
7. If you're thinking of unhosted wallets - that's correct. But a similar concern should be raised for the compliance key as well. If we're talking about custodial wallets, I don't think we should worry about it as the wallet private key will not be exposed to the end-user anyway. However, I did use different encryption algorithms with different key schemes to make such confusion harder.

lips/lip-9.md
danprinz: Agree. 
The main reason for the data is the user control over it. If we go off-chain, a salted hash would be enough.

lips/lip-9.md
danprinz: I agree that storing private data on-chain introduces a potential risk of having it exposed if a new crypto-attack will hit. 
Note that I don't suggest storing *personal* data by itself, rather the payment details (like description). the identity behind the data remains unknown to an outside observer.

lips/lip-9.md
```diff
@@ -0,0 +1,93 @@
...
```
sblackshear, line 25: One higher-level principle that I think is worth aligning on: what is the bar for adding a new kind of cryptographic key in Libra?

Each new key introduces a lot of complexity: someone needs securely store the private key, manage it (e.g., rotate regularly), publicize the public key, and (in this case) define a secure protocol for migrating keys from one vendor to another (if the user decides to switch). Another factor that should be taken into account is how many of these keys there will be--obviously, more keys = more complexity.

Right now (outside of validators), Libra only has account keys (one/account) and compliance keys (one/VASP). The motivation for each of these keys is quite clear--they buy a lot of use-cases. And the number of total keys is quite small--in the early days of Libra, there will probably be 10s-100s of VASPs and 100s-100000s of total accounts.

By contrast, the number of end-uses (i.e., wallet customers) will always be *much* higher than the number of VASPs or blockchain accounts. Thus, it feels like the bar for adding a new kind of cryptographic key that is (one/end-user) should be astronomically high. I don't think the utility provided by this LIP is sufficient motivation for adding this much complexity to Libra.

aching:
This is 5 months+ old, any objections to closing this request?  If there is a new suggestion for metadata that addresses the concerns, we could open a new DIP.

danprinz:
Agree. 
Closing it.

PR #81, opened 2020-10-31
Submitting lip-7

viktorbunin:
Lip-7 is a process lip on the lifecycle of technical working groups.

### DIFF
--- a/lips/lip-7.md
+++ b/lips/lip-7.md
@@ -1,0 +1,53 @@
+---
+lip: 7
+title: Technical Working Group Lifecycle
+authors: Viktor Bunin of Bison Trails (@viktorbunin)
+status: Draft
+type: Process
+created: 10/30/2020
+---
+
+
+## Abstract
+This LIP outlines the process by which technical working groups (TWG’s) are formed and how they should conduct their operations.
+
+## Formation of a Technical Working Group
+Technical Working Groups (TWG) are formed to address or explore specific technical topics or issues as they relate to Libra. TWGs collect information and expert input on a particular topic, discuss material changes to the Libra Payment Network (LPN), and create outputs in the form of recommendations, reports, and Libra Improvement Proposals (LIPs).
+
+The proposal to form a technical working group can be made by any Association member. The proposal must include, but is not limited to:
+The group’s scope and goal
+The limits of its focus
+The expected final output, typically as a LIP, report, or recommendation
+Expected duration of operation
+The initial facilitator (up to two individuals)
+A brief description of ideal desired member profiles so other Association members can join
+
+Once the proposal is created, the facilitator will circulate it in public slack channels relevant to the topic. The facilitator is also encouraged to proactively share the proposal with Association members that are likely to be impacted by this TWG or that could contribute to it. The facilitator must circulate the proposal over 1-2 weeks for sufficient interest to gather.
+
+After gathering interest, the facilitator must choose members to be part of the TWG, add them to the proposal, and identify their roles (co-facilitator, contributor, reviewer, etc.). The proposal must have representation from a minimum of 3 Association members to build collaboration and as a tool to determine relevance to the entire association, rather than an individual member.
+
+The facilitator is responsible for promoting and explaining the proposed TWG to Libra members and must lead it through the voting process by the Technical Steering Committee (TSC). They are also responsible for recruiting other members to the TWG as needed on an ongoing basis, and must help find a replacement facilitator if they must cease work on the initiative.
+
+Once the TWG proposal is presented to the TSC, a simple majority vote in favor forms the TWG. The TWG can then proceed in its stated work and make decisions as needed in the areas covered by its focus.
+
+## Decision Making in a TWG
+The TWG facilitator owns the agenda, meetings, and organization of the TWG, but is not a final decision maker. They are there to drive and facilitate discussion, and help allocate resources so that recommendations, reports, or draft LIPs can be created. They are also responsible for providing a bi-weekly update to the association on the progress the TWG is making and any associated documentation being produced. Bi-weekly updates should be posted in the Libra portal and distributed via email/slack to all Association members.
+
+If there are material changes to the TWG’s scope, timeline, or expected output, a revised proposal must be submitted to the TSC for re-approval.
+
+Any TWG member can propose any topic or idea within the scope of the TWG up for discussion with the group. These discussions should be documented, and as the group comes to agreement on how to proceed, there should be clear ownership among TWG members for drafting, editing, obtaining feedback, and finalizing the group’s working product. There does not need to be total agreement in the TWG on the final output, but a supermajority (67%+) vote in favor is required before proceeding.
+
+Once the TWGs final output is ready and the group votes to proceed, it must be shared with the TSC for review and edits. After the editing process, the TSC must approve the output before it can be circulated within the Association.
+
+If the output is a LIP, the TWG will submit the draft LIP as a pull-request to the libra/lips folder, entering the official [LIP process](https://lip.libra.org/overview/). At this point the LIP gets a LIP Manager to navigate the process from becoming a Draft LIP to a committed LIP in Github. 
+
+In addition to assisting with the LIP process, a LIP Manager can also dedicate engineering resources to develop an implementation of the LIP if one does not exist already. This may mean: 
+Libra Core changes (via github issues and PRs)
+Move Framework changes (same implementation as above)
+Non-blockchain changes in which case things like off-chain work may be affected
+
+After a LIP has sufficient engineering work or a full implementation ready, it enters the last call period. Before it can become Accepted, the TSC must vote on the LIP. A simple majority moves the LIP to Accepted, at which point it becomes up to the maintainers to implement it.
+
+## Example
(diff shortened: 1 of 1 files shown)


dahliamalkhi (approved):
Thanks for starting this !

I am approving the initial draft. However, I have some comments towards future revisions:
1. I am not sure we want hard requirements like "minimum 3 members" (how do you enforce this, take attendance?..), "push reports every 2 weeks" (some TWGs may be dormant at periods), etc. I'd prefer to think of these as guidelines for a successful/effective TWG.

2. I would add option to include participants who are not members. for example, Node Operations WG includes 3rd party operators, others groups may include experts, merchant representation, etc.

3. TWG may have more than a single LIP as output. For example, Off-chain Protocols is on-going, already produced 3 LIPs.

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+
+## Abstract
+This LIP outlines the process by which technical working groups (TWG’s) are formed and how they should conduct their operations.
+
+## Formation of a Technical Working Group
+Technical Working Groups (TWG) are formed to address or explore specific technical topics or issues as they relate to Libra. TWGs collect information and expert input on a particular topic, discuss material changes to the Libra Payment Network (LPN), and create outputs in the form of recommendations, reports, and Libra Improvement Proposals (LIPs).
+
+The proposal to form a technical working group can be made by any Association member. The proposal must include, but is not limited to:
```
kphfb, line 17: Do we want to close this off to only association members? I could see an argument for and against this.  Just curious what the reasoning was that you went with this model

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+The group’s scope and goal
+The limits of its focus
+The expected final output, typically as a LIP, report, or recommendation
+Expected duration of operation
+The initial facilitator (up to two individuals)
+A brief description of ideal desired member profiles so other Association members can join
+
+Once the proposal is created, the facilitator will circulate it in public slack channels relevant to the topic. The facilitator is also encouraged to proactively share the proposal with Association members that are likely to be impacted by this TWG or that could contribute to it. The facilitator must circulate the proposal over 1-2 weeks for sufficient interest to gather.
+
+After gathering interest, the facilitator must choose members to be part of the TWG, add them to the proposal, and identify their roles (co-facilitator, contributor, reviewer, etc.). The proposal must have representation from a minimum of 3 Association members to build collaboration and as a tool to determine relevance to the entire association, rather than an individual member.
```
kphfb, line 27: Why does the facilitator choose who will be part of the working group?  Is that to keep the groups small?  The reason I ask is that it could be odd if for example Bison Trails wants to join a working group but the facilitator says that they weren't chosen and thus can't participate in the group

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+
+The facilitator is responsible for promoting and explaining the proposed TWG to Libra members and must lead it through the voting process by the Technical Steering Committee (TSC). They are also responsible for recruiting other members to the TWG as needed on an ongoing basis, and must help find a replacement facilitator if they must cease work on the initiative.
+
+Once the TWG proposal is presented to the TSC, a simple majority vote in favor forms the TWG. The TWG can then proceed in its stated work and make decisions as needed in the areas covered by its focus.
+
+## Decision Making in a TWG
+The TWG facilitator owns the agenda, meetings, and organization of the TWG, but is not a final decision maker. They are there to drive and facilitate discussion, and help allocate resources so that recommendations, reports, or draft LIPs can be created. They are also responsible for providing a bi-weekly update to the association on the progress the TWG is making and any associated documentation being produced. Bi-weekly updates should be posted in the Libra portal and distributed via email/slack to all Association members.
```
kphfb, line 34: I'd echo Dahlia's comment that this should be a suggested practice rather than "you must post every 2 weeks"

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+
+If there are material changes to the TWG’s scope, timeline, or expected output, a revised proposal must be submitted to the TSC for re-approval.
```
kphfb, line 36: The timeline one is a bit strange to me.  If a working group isn't making progress quickly enough, now they need to be re-approved by the TSC to continue their work?

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+
+Any TWG member can propose any topic or idea within the scope of the TWG up for discussion with the group. These discussions should be documented, and as the group comes to agreement on how to proceed, there should be clear ownership among TWG members for drafting, editing, obtaining feedback, and finalizing the group’s working product. There does not need to be total agreement in the TWG on the final output, but a supermajority (67%+) vote in favor is required before proceeding.
+
+Once the TWGs final output is ready and the group votes to proceed, it must be shared with the TSC for review and edits. After the editing process, the TSC must approve the output before it can be circulated within the Association.
```
kphfb, line 40: "After the editing process, the TSC must approve the output before it can be circulated within the Association" - it's unclear what is being suggested here.  By creating a draft PR, are we not effectively "circulating within the Association"?

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+
+If the output is a LIP, the TWG will submit the draft LIP as a pull-request to the libra/lips folder, entering the official [LIP process](https://lip.libra.org/overview/). At this point the LIP gets a LIP Manager to navigate the process from becoming a Draft LIP to a committed LIP in Github. 
+
+In addition to assisting with the LIP process, a LIP Manager can also dedicate engineering resources to develop an implementation of the LIP if one does not exist already. This may mean:
```
kphfb, line 44: Do we want to call out that a LIP manager can dedicate eng resources?  Doesn't that hold true to anyone (even if not involved in the TWG)?

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+Once the TWGs final output is ready and the group votes to proceed, it must be shared with the TSC for review and edits. After the editing process, the TSC must approve the output before it can be circulated within the Association.
+
+If the output is a LIP, the TWG will submit the draft LIP as a pull-request to the libra/lips folder, entering the official [LIP process](https://lip.libra.org/overview/). At this point the LIP gets a LIP Manager to navigate the process from becoming a Draft LIP to a committed LIP in Github. 
+
+In addition to assisting with the LIP process, a LIP Manager can also dedicate engineering resources to develop an implementation of the LIP if one does not exist already. This may mean: 
+Libra Core changes (via github issues and PRs)
+Move Framework changes (same implementation as above)
+Non-blockchain changes in which case things like off-chain work may be affected
+
+After a LIP has sufficient engineering work or a full implementation ready, it enters the last call period. Before it can become Accepted, the TSC must vote on the LIP. A simple majority moves the LIP to Accepted, at which point it becomes up to the maintainers to implement it.
```
kphfb, line 49: > at which point it becomes up to the maintainers to implement it

I'm not sure that it becomes up to the maintainers to implement it.  Maintainers may not have the bandwidth to fully implement it and may require additional resources from other Association members

lips/lip-7.md
```diff
@@ -0,0 +1,53 @@
...
+
+Once the TWGs final output is ready and the group votes to proceed, it must be shared with the TSC for review and edits. After the editing process, the TSC must approve the output before it can be circulated within the Association.
+
+If the output is a LIP, the TWG will submit the draft LIP as a pull-request to the libra/lips folder, entering the official [LIP process](https://lip.libra.org/overview/). At this point the LIP gets a LIP Manager to navigate the process from becoming a Draft LIP to a committed LIP in Github. 
+
+In addition to assisting with the LIP process, a LIP Manager can also dedicate engineering resources to develop an implementation of the LIP if one does not exist already. This may mean: 
+Libra Core changes (via github issues and PRs)
+Move Framework changes (same implementation as above)
```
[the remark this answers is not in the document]
viktorbunin: Updating, thanks! (to Diem)

lips/lip-7.md
viktorbunin: This is a great point. I was thinking originally that it would just be association members, but I don't see a truly stellar reason to restrict it. If for some reason there is a problem with having it be open, it can always be restricted in the future. Updating this!

lips/lip-7.md
viktorbunin: This is a great point. The intention here was to have the facilitator own drumming up support and interest, but not to be exclusionary. I will update to make it clear that membership is open to all. Thank you for pointing it out!

lips/lip-7.md
viktorbunin: Updated, thank you!

lips/lip-7.md
viktorbunin: Fair point. I also removed expected output because that can also change over time. Really the only thing we need to watch out for is material changes in scope because we shouldn't have a TWG get approval to work on one thing and then pivot to something completely different without getting approval for the new direction.

lips/lip-7.md
viktorbunin: Sorry, but I'm not understanding your question. Can you say it a different way?

lips/lip-7.md
viktorbunin: Updated to majority. Also by circulating, I meant that it gets the stamp of approval from the TSC as a recommendation. I'll update to make that more clear.

PR #82, opened 2020-11-03
[Lip 10] Libra ID Spec

andll:
# Open questions

## Libra ID separator
Contenders: @, # and $
“@”:
(-) Too easy to confuse with an email
“$”:
(+/-) Already used by Ripple’s Pay ID
(+/-) Can be seen as an affiliation with USD. However, it is fairly common to use $ as a sign for “money”, rather than a specific currency, so this can also be a plus.
(+) Can be used unencoded in the URLs
“#”:
(-) If placed into URL strictly requires a URL encoding

## Libra ID domain - single word or a domain name
Single word(e.g. andrey$novi):
(+) Fewer symbols to write
Domain name(e.g. andrey$novi.com):
(+) In the future if we allow unhosted wallets, we can switch to DNS lookups
(+) For the merchant use case, can create a merchant APP url using string manipulation

## Profile picture sizes and image formats
We can put some limits on maximum file size and image resolution.


andll:
cc @gdanezis @danprinz - can assign you as reviewers, but please take a look
@danprinz Can you also cc more people from first to this LIP(don't know their github user names)

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
+---
+lip: 10
+title: Libra ID Spec
+authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb)
+status: Draft
+type: Informational
```
kphfb, line 6: @dahliamalkhi I'm guessing this should be a Standard LIP?

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+# Summary
+
+This LIP describes Libra ID - the human-readable identifier for user accounts.   
+
+In addition to describing the format of the Libra ID, this LIP also describes the off-chain protocol for the peer-2-peer payment with Libra ID.
+
+This document does not describe an off-chain protocol for using Libra ID for merchant use case - this protocol will be part of merchant flow LIP.  
+
+# Motivation
+Libra ID provides a convenient way to identify user's account within a VASP. There are two main use cases for Libra ID:
```
kphfb, line 18:
```suggestion
Libra ID provides a convenient way to identify a user's account within a VASP. There are two main use cases for Libra ID:
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+* For peer-2-peer payments, Libra ID allows a user to provide a human-readable ‘address’ to the sender. Basically, instead of sending a lip-5 address(that needs to be refreshed for each payment, ideally), the receiver can just share their Libra ID once. In this case Libra ID plays the role of an email address for payments. Benefits for this use case:
+  * Libra ID is a user-friendly identifier, unlike lip-5 address
```
kphfb, line 20:
```suggestion
  * Libra ID is a user-friendly identifier, unlike lip-5 addresses
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+This LIP describes Libra ID - the human-readable identifier for user accounts.   
+
+In addition to describing the format of the Libra ID, this LIP also describes the off-chain protocol for the peer-2-peer payment with Libra ID.
+
+This document does not describe an off-chain protocol for using Libra ID for merchant use case - this protocol will be part of merchant flow LIP.  
+
+# Motivation
+Libra ID provides a convenient way to identify user's account within a VASP. There are two main use cases for Libra ID:
+* For peer-2-peer payments, Libra ID allows a user to provide a human-readable ‘address’ to the sender. Basically, instead of sending a lip-5 address(that needs to be refreshed for each payment, ideally), the receiver can just share their Libra ID once. In this case Libra ID plays the role of an email address for payments. Benefits for this use case:
```
kphfb, line 19: Let's add a link to LIP 5

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+  * Libra ID is a user-friendly identifier, unlike lip-5 address
+  * Libra ID resolution process can allow to exchange information about both sender and receiver for payment
```
kphfb, line 21:
```suggestion
  * The Libra ID resolution process can facilitate the exchange of rich information (e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+  * Libra ID resolution allows to fetch details about each participant, like profile picture
```
kphfb, line 22: I think this one is basically the same as above unless I'm misunderstanding the above

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+  * Libra ID allows for better anonymity on-chain, as it is possible to share it once, but for wallet apps to transparently resolve it to new subaddress for each transaction. This is a big improvement comparing to exchange of lip-5 identifiers, that leaks privacy information on the chain and allows to track all transactions for the individual subaccount.
```
kphfb, line 23:
```suggestion
  * While it is encouraged for VASPs to treat subaddresses as single-use currently, Libra ID allows for this anonymity on-chain to be handled in a much simpler manner.  LibraID can be shared once with wallet apps transparently resolving the LibraID into a new identifier for each transaction (and thereby preventing linkability which could happen if subaddresses were re-used).
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+* For the merchant use case, Libra ID allows users to self-identify themself to merchants. In this case Libra ID plays the role of credit card number or also can be seen as a variant of OpenID-type protocol.
```
kphfb, line 24:
```suggestion
* For the merchant use case, Libra ID allows users to self-identify to merchants. In this case, Libra ID plays the role of credit card number or also can be seen as a variant of OpenID-type protocol.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Examples in other blockchains
+Ripple proposes PayID. PayID has some similarity to the current proposal, but is not entirely the same thing.
```
kphfb, line 27:
```suggestion
Ripple proposes PayID. PayID has some similarity to the current proposal, but is not entirely identical.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+For example, PayID seems to target cross-chain payments, while libra id is supposed to facilitate cross-vasp payments, but only inside the libra network.
```
kphfb, line 28:
```suggestion
For example, PayID seems to target cross-chain payments, while libra id is intended to facilitate cross-vasp payments, but only inside the libra network.
```

kphfb, line 28: I wonder if we want to mention that the initial target for LibraID is within Libra, but that nothing in the design specifically prevents cross-chain.  And in fact, we should actually think through to make sure that is the case.  I could see it potentially being useful to go across chains at some point.  Maybe the lookup still needs to happen on Libra, but in the off-chain messages, maybe we can later add a chain identifier that coudl be eth or bitcoin, etc.  I don't think this is something we should build now, but I wonder if we can feel pretty certain that this would be possible in the future if we later desired to do so

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+# Libra ID format
+Libra ID is a string in format `<user-id>$<libra-id-domain>`.
+
+* **user-id**, identifies an account within a VASP:
+  * This part is case insensitive.
```
kphfb, line 34:
```suggestion
  * Case insensitive.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]?$`
+  * This part is not published on the chain, there is no way for an external entity to list all the users within VASP.
```
kphfb, line 36:
```suggestion
  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+* **libra-id-domain** identifies VASP:
+  * This part is case insensitive.
```
kphfb, line 38:
```suggestion
  * Case insensitive.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]?$`
+  * libra-id-domains are published on the chain
+  * Each libra-id-domain is uniquely associated with single VASP
+  * VASP can have multiple libra-id-domains associated, thought it is not expected to be common
```
kphfb, line 42:
```suggestion
  * VASPs can have multiple libra-id-domains associated, though it is not expected to be common
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Multiple accounts per user
+VASP should not assume that Libra ID uniquely identifies the user within the system. User that have accounts with different VASP would have different Libra ID within each VASP.
```
kphfb, line 45:
```suggestion
VASPs should not assume that Libra ID uniquely identifies the user within the system. Users that have accounts with different VASPs would have different Libra ID within each VASP.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+Furthermore, a user can have multiple Libra IDs within the same VASP. This does not affect the protocol, but VASP can leverage this, for example to allow users to create multiple ‘accounts’ like checking, savings, etc.
```
kphfb, line 47:
```suggestion
Furthermore, a user can have multiple Libra IDs within the same VASP. This does not affect the protocol, but VASPs can leverage this - for example to allow users to create multiple ‘accounts’ like checking, savings, etc.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+VASP can also allow users to generate temporary / situational Libra IDs. This can be used in many scenarios such as:
```
kphfb, line 49:
```suggestion
VASPs can also allow users to generate temporary / situational Libra IDs. This can be used in many scenarios such as:
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+Each libra ID domain is associated with a VASP. The source of truth for this association is a blockchain - a special LibraIdDomains resource must be published into VASP for it to expose it’s LibraID domain.
+
+```
+resource struct LibraIdDomains {
+    domains: vector<LibraIdDomain>,
+}
+
+struct LibraIdDomain {
+    domain: vector<u8>,
+    wallet_uri: vector<u8>,
```
kphfb, line 64: Why vec<u8> instead of strings? Or at least what encoding are the bytes?  ASCII?

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+}
+```
+
+* The `LibraIdDomains` resource can only be published into an account with `Role == PARENT_VASP_ROLE_ID`.
+* The `LibraIdDomains` contains a list of `LibraIdDomain` structs that can be associated with a VASP. As such It is possible to register more than one Libra ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP on the chain.
```
kphfb, line 69:
```suggestion
* The `LibraIdDomains` contains a list of `LibraIdDomain` structs that can be associated with a VASP. As such, it is possible to register more than one Libra ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+* Only special Treasury Compliance account(address `0xB1E55D`) can manipulate LibraIdDomains resource:
+  * Only TC account can create and publish LibraIdDomains resource
+  * Only TC account can add, remove or update an LibraIdDomain within LibraIdDomains resource
+* In order to register a Libra Id domain, VASP needs to submit a request to Libra Association. LA will perform certain checks(out of scope of this document) before they will issue an on-chain transaction to register a Libra Id Domain. The checks that LA will perform will make sure that VASPs do not claim irrelevant Libra Id Domains. LA submission will also ensure that Libra Id Domains are unique.
+* It should also be possible(and probably going to be most common) for VASP to submit their desired Libra ID domain(s) during the registration.
```
kphfb, line 74:
```suggestion
* It should also be possible(and probably going to be most common) for VASPs to submit their desired Libra ID domain(s) during the registration.
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+## UI flow
+* Receiver shares their Libra ID with sender
+* Sender enters receiver’s Libra ID(and other txn details such as amount) in their wallet app
+* Sender’s wallet can fetch profile picture and full name of the receiver prior to sending money, and show it to sender, to improve security (if receiver VASP allows it)
+* Sender presses ‘send’ button
+* Receiver receives money and information about the sender (optionally)
+
+## Backend flow
+In order to support peer-2-peer payments, VASPs need to implement two Libra ID endpoints:
+* **Info endpoint** for querying optional information about the receiver. VASPs can provide this endpoint to improve UX and security of payment, but they are not obligated to - VASP may choose not to share any information about the user via this endpoint, including whether the user exists or not
```
kphfb, line 89:
```suggestion
* **Info endpoint** for querying optional information about the receiver. VASPs can provide this endpoint to improve UX and security of payment, but they are not obligated to - VASPs may choose not to share any information about the user via this endpoint, including whether the user exists or not
```

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+* **Payment endpoint** to perform actual money transfer.
+
+Common considerations:
+
+* For both endpoints, all communications are signed via the same mechanism as for other off-chain communications, see the Payloads section of lip-1 for details. As such, all vasp-2-vasp communications are authenticated.
```
kphfb, line 94: Let's add a link to LIP 1

lips/lip-10.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Info endpoint
+
+```
+POST https://<hostname>:<port>/pay.<protocol_version>/info
+```
+
+**Info endpoint** is used by the **Sender** to request information about the **Receiver** before submitting an actual payment.
+
+Info endpoint takes bother sender and receiver Libra IDs as arguments. Receiver VASP might decide to share more(or less) information based on sender ID, for example if the wallet supports some kind of contact list, they might decide to share more information if the sender is in the contact list of the receiver.
```
kphfb, line 112:
```suggestion
Info endpoint takes both sender and receiver Libra IDs as arguments. Receiver VASP might decide to share more(or less) information based on sender ID, for example if the wallet supports some kind of contact list, they might decide to share more information if the sender is in the contact list of the receiver.
```

kphfb (commented):
Mostly just small nits.  Will come back and review the meat of the protocol

lips/lip-10.md
andll: Afaik there is simply no string type in move

andll: Added clarification on the encoding

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+**Info endpoint** is used by the **Sender** to request information about the **Receiver** before submitting an actual payment.
+
+Info endpoint takes both sender and receiver Libra IDs as arguments. Receiver VASP might decide to share more(or less) information based on sender ID, for example if the wallet supports some kind of contact list, they might decide to share more information if the sender is in the contact list of the receiver.
+
+**Request:**
+```json
+{
+   "sender_id": "<Sender Libra ID>",
```
kphfb, line 116: I wonder if we should allow the sender to optionally send additional info such as the user object. It removes the request the other direction to fetch that info and maybe a receiver is more likely to show such information like you mention if they are in their contact list.  So if the sender says they have phone number 123-4567 and that's in the user's contact book, maybe they are more likely to share info?  Not sure if this is a good idea or not, just something to think about

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+
+**Response:**
+```json
+{
+   "receiver": <User Object>
+}
+```
+
+* Receiver Libra ID domain should belong to the Receiver VASP. Otherwise, Receiver VASP should return the `400` error code.
+* Sender Libra ID should have a Libra ID domain that belongs to Sender VASP. If receiver VASP makes a decision on how much information to share, based on the sender Libra ID, it should enforce this property and return 400 error code, if VASP tries to impersonate a user from Libra ID domain that it does not own.
```
kphfb, line 129:
```suggestion
* Sender Libra ID must have a Libra ID domain that belongs to Sender VASP. If this is not the case, the Receiver VASP should return the `400` error code 
* Receiver VASP may optionally choose how much information to share based on the sender Libra ID. If Receiver VASP decides to not share data, it should return `400` error code.
```

kphfb: I edited it to split it out for how I believe you intended it to read, but I actually think that we shouldn't return the same error code for "i don't want to share information" that we do for "this person doesn't exist".  In one of those cases, you can still choose to send money to them

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Pay endpoint
```
kphfb, line 131: I think we should structure this in a similar way to what is done here: https://lip.libra.org/lip-1/#commandresponseobject

Where we specify the types of each field, whether it's optional, a short description of it, etc.

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+
+**Pay endpoint** is used when a user is willing to make a money transfer.
+
+**Request:**
+```json
+{
+   "sender": <User Object>,
+   "receiver_id": "<Receiver_Libra_ID>",
+   "amount": "<amount>",
+   "currency": "<currency>"
```
kphfb, line 145: Do you think these are needed?  For example, what if a VASP pre-fetches this info before their user has filled out the payment amounts?

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+}
+```
+
+**Response:**
+```json
+{
+   "receiver_address": "<Receiver VASP on-chain account address>",
+   "reference_id": "<Payment_Reference_ID>",
```
kphfb, line 153: Hmmm, as I look at it more, I do wonder if we actually need two endpoints.  It feels like this is really just one endpoint that has a flag for if you want a reference ID or not.  Otherwise, they are effectively the same thing

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+   "receiver_id": "<Receiver_Libra_ID>",
+   "amount": "<amount>",
+   "currency": "<currency>"
+}
+```
+
+**Response:**
+```json
+{
+   "receiver_address": "<Receiver VASP on-chain account address>",
```
kphfb, line 152: Need to specify for this and all fields more info about them.  Is this bech32?

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+   "full_name": "<Full name>",
+   "profile_picture": "<profile_picture_uri>"
+}
+```
+If a user does not exist, VASP can choose to return an HTTP error, or the empty User Object, depending on it’s privacy choices.
+## Profile Picture URI
+Profile picture URI is a URI that represents a profile picture of a user. Wallet app should support two possible URI schemas for profile picture.
+
+* Https schema. If URI starts with https:, wallet should attempt to fetch profile picture from the HTTP URI provided. When fetching this URI wallet app will not provide any authentication. Wallet app will use GET method to fetch the profile picture specified using https schema.
+* Data schema. If URI starts with data:, wallet should interpret the URI as a Data URL that contains profile picture encoded.
```
kphfb, line 187: We should show examples of each of these

lips/lip-10.md
```diff
@@ -0,0 +1,190 @@
...
+One of the main use cases for Libra ID is transferring money from one user to another across different wallets.
+
+## UI flow
+* Receiver shares their Libra ID with sender
+* Sender enters receiver’s Libra ID(and other txn details such as amount) in their wallet app
+* Sender’s wallet can fetch profile picture and full name of the receiver prior to sending money, and show it to sender, to improve security (if receiver VASP allows it)
+* Sender presses ‘send’ button
+* Receiver receives money and information about the sender (optionally)
+
+## Backend flow
```
kphfb, line 86: I think it would be good to have a diagram demonstrating how this flow will work in a quickly parsable format

lips/lip-10.md
andll: For prefetching, VASP should use info endpoint, this is payment endpoint, so I think currency / amount is needed

lips/lip-10.md
andll: To me two separate endpoints makes more sense. Yes they share some fields (like there is sender/receiver), but there are also many fields that are not shared (info endpoint don't have amounts/coin in request, and does not return reference/account address)
They also have different semantics - info endpoint is 'read' endpoint, while 'pay' is a 'write' endpoint as it needs to create an entry in the database

lips/lip-10.md
andll: Tried to rephrase it. I think if VASP don't want to share info, but user exists, they can just return empty user object(e.g. not sharing any fields). Otherwise return error

andll:
cc @bmaurer

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+
+# Libra ID format
+Libra ID is a string in format `<user-id>$<libra-id-domain>`.
+
+* **user-id**, identifies an account within a VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]?$`
+  * Maximum length: 64 characters
+  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
+* **libra-id-domain** identifies the VASP:
```
bmaurer, line 40: Is this a domain as in libra.org? or is it not connected to the domain name system

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+
+Common considerations:
+
+* For both endpoints, all communications are signed via the same mechanism as for other off-chain communications, see the [Payloads section of lip-1](https://github.com/libra/lip/blob/master/lips/lip-1.mdx#payloads) for details. As such, all vasp-2-vasp communications are authenticated.
+
+* All Libra ID endpoints are nested under `pay.<version>` URI path:
+
+     `https://<hostname>:<port>/pay.<protocol_version>/<endpoint>`
+
+* All Libra ID endpoints use the same `base_url` and same `compliance_public_key` for authentication as other off-chain APIs.
```
bmaurer, line 108: It might make sense to break out standards for APIs into its own LIP. This is lacking a bit of specificity.

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+
+| Field 	      | Type  | Required? 	| Description 	           |
+|-------	      |------ |-----------	|-------------	           |
+|libra_id         |str    | Y           | Libra ID of the user |
+|display_name     |str    | N           | Display name of the user. Usually a full name |
+|profile_picture  |str    | N           | Profile picture URL(see below) |
+
+If a user does not exist, VASP can choose to return an HTTP error, or the empty User Object, depending on it’s privacy choices.
+
+## Profile Picture URI
```
bmaurer, line 225: May make sense to discuss privacy implications (e.g., if you just hot link to another VASP that VASP can track how often the profile pic is accessed). Generally at least for FB we have it as a best practice that you always proxy this type of access.

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+* **libra-id-domain** identifies the VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]?$`
+  * Maximum length: 63 characters
+  * libra-id-domains are published on the chain
+  * Each libra-id-domain is uniquely associated with single VASP
+  * VASPs can have multiple libra-id-domains associated, though it is not expected to be common
+
+## Multiple accounts per user
+VASPs should not assume that Libra ID uniquely identifies the user within the system. Users that have accounts with different VASPs would have different Libra ID within each VASP.
```
bmaurer, line 49: MUST NOT? (this happens a lot through this document... not going to audit each instance. I think it's worth doing a pass to audit the use of MUST/SHOULD/MAY)

bmaurer:
At a higher level: it feels like we have a lot of overlap with the off-chain protocol / subaddresses. For example:

* If you have a subaddress you can use the off-chain protocol to get a reference ID from the subaddress (and submit travel rule info)
* If you have a libra ID, you first get a reference ID, then use the off-chain protocol with the reference ID to submit more information.

I worry that we start to get complexity from different permutations of these APIs.

Is there anything we can do to be more uniform here? Both subaddresses and LibraID are a way to reference a user within a VASP. Why entirely different flows for the two methods?

It's OK if we have regrets about APIs we designed for subaddresses and want to go a different direction. But in that case, can we come up with a new API that is unifying vs one that only addresses a subset of the use cases.

lips/lip-10.md
andll: This is somewhat of an open question, see 'Libra ID domain - single word or a domain name' section in the PR description.

In it's current from this spec does not specify connection between libra-id-domain and DNS.

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+* Sender enters receiver’s Libra ID(and other txn details such as amount) in their wallet app
+* Sender’s wallet can fetch profile picture and full name of the receiver prior to sending money, and show it to sender, to improve security (if receiver VASP allows it)
+* Sender presses ‘send’ button
+* Receiver receives money and information about the sender (optionally)
+
+## Backend flow
+In order to support peer-2-peer payments, VASPs need to implement two Libra ID endpoints:
+
+
+* **Info endpoint** for querying optional information about the receiver. VASPs can provide this endpoint to improve UX and security of payment, but they are not obligated to - VASPs may choose not to share any information about the user via this endpoint, including whether the user exists or not
```
sblackshear, line 97: How will they make the decision not to share information?

lips/lip-10.md
sblackshear, line 108: Does using `compliance_public_key` make sense here? We need it for for TR because the signature will be checked on-chain, but LibraID seems like it could use a more conventional authentication mechanism. Not my area of expertise at all, but maybe someone who knows a lot about web authentication could weigh in.

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+* Only special Treasury Compliance account(address `0xB1E55D`) can manipulate LibraIdDomains resource:
+  * Only TC account can create and publish LibraIdDomains resource
+  * Only TC account can add, remove or update an LibraIdDomain within LibraIdDomains resource
+* In order to register a Libra Id domain, VASP needs to submit a request to Libra Association. LA will perform certain checks(out of scope of this document) before they will issue an on-chain transaction to register a Libra Id Domain. The checks that LA will perform will make sure that VASPs do not claim irrelevant Libra Id Domains. LA submission will also ensure that Libra Id Domains are unique.
+* It should also be possible(and probably going to be most common) for VASPs to submit their desired Libra ID domain(s) during the registration.
```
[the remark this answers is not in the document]
andll: We don't have doc for this process yet, detailed runbook for association is probably too early to figure out

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]?$`
+  * Maximum length: 64 characters
+  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
+* **libra-id-domain** identifies the VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]?$`
+  * Maximum length: 63 characters
+  * libra-id-domains are published on the chain
+  * Each libra-id-domain is uniquely associated with single VASP
+  * VASPs can have multiple libra-id-domains associated, though it is not expected to be common
```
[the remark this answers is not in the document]
andll: There is actually more details and motivation below when discussing on-chain format and restrictions there. I am going to remove this item and keep this section focused on high level overview of ID structure, to avoid duplication and confusion

lips/lip-10.md
andll: Yeah, I think currently `compliance_public_key` is simply overloaded - it is used to both signing KYC metadata, _and_ for signing the requests. However, this is how things are set in lip-1(main offchain lip), and this document just re-states it for clarity

lips/lip-10.md
andll: Added example inline

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+  * Libra ID is a user-friendly identifier, unlike lip-5 addresses
+  * The Libra ID resolution process can facilitate the exchange of rich information(e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
+  * While it is encouraged for VASPs to treat subaddresses as single-use currently, Libra ID allows for this anonymity on-chain to be handled in a much simpler manner.  LibraID can be shared once with wallet apps transparently resolving the LibraID into a new identifier for each transaction (and thereby preventing linkability which could happen if subaddresses were re-used).
+* For the merchant use case, Libra ID allows users to self-identify to merchants. In this case, Libra ID plays the role of credit card number or also can be seen as a variant of OpenID-type protocol.
```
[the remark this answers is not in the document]
andll: * There is _some_ analogy in a sense that you enter your cc number in order to get 'routed' to the bank
* I wanted to cover it in merchant flow doc, but tldr is - yes, you need login into wallet app / web site to use your libra ID

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+    domain: vector<u8>,     // Utf-8 encoding
+    wallet_uri: vector<u8>, // ASCII encoding (use url-escape for non-ASCII characters)
+}
+```
+
+* The `LibraIdDomains` resource can only be published into an account with `Role == PARENT_VASP_ROLE_ID`.
+* The `LibraIdDomains` contains a list of `LibraIdDomain` structs that can be associated with a VASP. As such, it is possible to register more than one Libra ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
+* Only special Treasury Compliance account(address `0xB1E55D`) can manipulate LibraIdDomains resource:
+  * Only TC account can create and publish LibraIdDomains resource
+  * Only TC account can add, remove or update an LibraIdDomain within LibraIdDomains resource
```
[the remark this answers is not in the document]
andll: I think since LA is allocating those 'domains' to the VASPs, there is no really problem with squatting, similarly all the questions about what happens to removed domain, etc, can be resolved by LA when specific case is reviewed. Do you think we need to include this details here?

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+Each libra ID domain is associated with a VASP. The source of truth for this association is a blockchain - a special LibraIdDomains resource must be published into VASP for it to expose it’s LibraID domain.
+
+```
+resource struct LibraIdDomains {
+    domains: vector<LibraIdDomain>,
+}
+
+struct LibraIdDomain {
+    domain: vector<u8>,     // Utf-8 encoding
+    wallet_uri: vector<u8>, // ASCII encoding (use url-escape for non-ASCII characters)
```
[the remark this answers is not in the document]
andll: Yeah.
Well, another option is this - I can say that wallet_uri is utf-8 encoded, but you still only allow to use ASCII characters and use escaping for everything else. But saying that 'it is utf-8 encoded with only ASCII characters allowed' seem to be a very convoluted way to say 'encoding is ASCII'

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+}
+```
+
+* The `LibraIdDomains` resource can only be published into an account with `Role == PARENT_VASP_ROLE_ID`.
+* The `LibraIdDomains` contains a list of `LibraIdDomain` structs that can be associated with a VASP. As such, it is possible to register more than one Libra ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
```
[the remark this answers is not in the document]
andll: Probably don't need this in standard, but there will be some natural limitations like account size. In practice I think it will be so rare to have more then one domain per VASP, that we probably don't need to have it as part of the standard?

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
+* **libra-id-domain** identifies the VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]?$`
+  * Maximum length: 63 characters
+  * libra-id-domains are published on the chain
+  * Each libra-id-domain is uniquely associated with single VASP
+  * VASPs can have multiple libra-id-domains associated, though it is not expected to be common
+
+## Multiple accounts per user
```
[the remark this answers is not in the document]
andll: So main motivation to this question is because some people might ask 'is libra id identifying a user, or specific account of the user". So this section is trying to clarify that. Do you think i need to add some clarification for it?

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+
+# Motivation
+Libra ID provides a convenient way to identify a user's account within a VASP. There are two main use cases for Libra ID:
+* For peer-2-peer payments, Libra ID allows a user to provide a human-readable ‘address’ to the sender. Basically, instead of sending a [lip-5 address](https://github.com/libra/lip/blob/master/lips/lip-5.md) (that needs to be refreshed for each payment, ideally), the receiver can just share their Libra ID once. In this case Libra ID plays the role of an email address for payments. Benefits for this use case:
+  * Libra ID is a user-friendly identifier, unlike lip-5 addresses
+  * The Libra ID resolution process can facilitate the exchange of rich information(e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
+  * While it is encouraged for VASPs to treat subaddresses as single-use currently, Libra ID allows for this anonymity on-chain to be handled in a much simpler manner.  LibraID can be shared once with wallet apps transparently resolving the LibraID into a new identifier for each transaction (and thereby preventing linkability which could happen if subaddresses were re-used).
```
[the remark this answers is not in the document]
andll: I changed wording there, let me know what you think

lips/lip-10.md
```diff
@@ -0,0 +1,234 @@
...
+* For the merchant use case, Libra ID allows users to self-identify to merchants. In this case, Libra ID plays the role of credit card number or also can be seen as a variant of OpenID-type protocol.
+
+## Examples in other blockchains
+Ripple proposes PayID. PayID has some similarity to the current proposal, but is not entirely identical.
```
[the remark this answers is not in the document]
andll: After some thought I removed PayID section from the doc. We spent some time researching it, however perhaps it is safer to just not discuss this as part of the LIP

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+  * Libra ID provides better privacy, comparing to using sub-addresses, because Libra ID is not shared on the chain
+  * Libra ID is a persistent user identifier. By contrast, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy.
+* For the merchant use case, Libra ID allows users to self-identify to merchants. In this case, Libra ID plays the role of credit card number or also can be seen as a variant of OpenID-type protocol.
+
+# Libra ID format
+Libra ID is a string in format `<user-id>$<libra-id-domain>`.
+
+* **user-id**, identifies an account within a VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]?$`
```
bob-wilson, line 34: That regexp looks wrong. I think you want a "*" instead of a "?". Otherwise, the user-ids can only be 1 or 2 characters in length.

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+  * Maximum length: 64 characters
+  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
+* **libra-id-domain** identifies the VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]?$`
```
bob-wilson, line 39: Same thing as above: "*" instead of "?"

lips/lip-10.md
andll: Good call!

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+     `https://<hostname>:<port>/pay.<protocol_version>/<endpoint>`
+
+* All Libra ID endpoints use the same `base_url` and same `compliance_public_key` for authentication as other off-chain APIs.
+
+* All Libra ID endpoints are HTTP POST endpoints, they accept JSON-encoded requests and return JSON-encoded responses. Other encoding methods can be considered in the future, `Content-Type` and `Accept` HTTP headers can be used to define request/response formats.
+
+## Info endpoint
+
+```
+POST https://<hostname>:<port>/pay.<protocol_version>/info
```
danprinz, line 112: I think that using OpenAPI (3) spec would be a more convenient way to describe the API and also could be used in a wide range of web frameworks

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+| Field 	   | Type 	     | Required? 	| Description 	             |
+|-------	   |------	     |-----------	|-------------	             |
+|receiver      |User Object  | Y            | Information about the receiver |
+
+* Receiver Libra ID domain must belong to the Receiver VASP. Otherwise, Receiver VASP must return the `400` error code.
+* Sender Libra ID must have a Libra ID domain that belongs to Sender VASP. If this is not the case, the Receiver VASP should return the `400` error code 
+* Receiver VASP may optionally choose how much information to share based on the sender Libra ID. If Receiver VASP decides to not share data, it should return an empty `User Object` with only `libra_id` field set
+* If user is not found, the VASP should return error code `400`
```
danprinz, line 146: Why not sticking to HTTP standards and reply with 404? 
400 should be a signal of a bad request, and I would argue that if the target user is not found doesn't mean an error or a bad request

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+* Receiver Libra ID domain must belong to the Receiver VASP. Otherwise, Receiver VASP must return the `400` error code.
+* Sender Libra ID must have a Libra ID domain that belongs to Sender VASP. If this is not the case, the Receiver VASP should return the `400` error code 
+* Receiver VASP may optionally choose how much information to share based on the sender Libra ID. If Receiver VASP decides to not share data, it should return an empty `User Object` with only `libra_id` field set
+* If user is not found, the VASP should return error code `400`  
+
+## Pay endpoint
+
+```
+POST https://<hostname>:<port>/pay.<protocol_version>/pay
```
danprinz, line 151: This is a "proper" off-chain command and should be treated as such. We should aim to have the reasoning behind the URI scheme. (see above)

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+created: 11/03/2020
+---
+
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+# Summary
+
+This LIP describes Libra ID - the human-readable identifier for user accounts.   
+
+In addition to describing the format of the Libra ID, this LIP also describes the off-chain protocol for the peer-2-peer payment with Libra ID.
```
danprinz, line 16: What is the main scenario that requires the off-chain p2p payment protocol?
For an amount of more than 1000, we must use TR metadata and signature. 
Is this a "nice-to-have" protocol for payments? any enforcement for that to take place before the actual transaction submission?

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+  * *For example*: VASP can have notion of 'contact list' and only share information if the sender is in receiver's contacts
+* **Payment endpoint** to perform actual money transfer.
+
+Common considerations:
+
+* For both endpoints, all communications are signed via the same mechanism as for other off-chain communications, see the [Payloads section of lip-1](https://github.com/libra/lip/blob/master/lips/lip-1.mdx#payloads) for details. As such, all vasp-2-vasp communications are authenticated.
+
+* All Libra ID endpoints are nested under `pay.<version>` URI path:
+
+     `https://<hostname>:<port>/pay.<protocol_version>/<endpoint>`
```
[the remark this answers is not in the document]
andll: I think I agree with this, I will change URL to make it more uniform

lips/lip-10.md
andll: Are you talking about adding `<LocalVASPAddress>/<RemoteVASPAddress>` part (agree with that) or something else?

lips/lip-10.md
andll: With Libra ID in current form any p2p payment requires off-chain first, since there is no subadresses when Libra ID is used

We were considering having a shortcut protocol (for example including encrypted libra id in the metadata) that would not require off-chain communication. But so far it feels like it does not simplify things that much(since the off-chain API for libra id is pretty simple POST request).

lips/lip-10.md
danprinz: both the `<LocalVASPAddress>/<RemoteVASPAddress>` part and the `pay` resource name. 
In general, LIP-1 (and LIP-8 as its extension) uses a single resource URI (`/command`) as the main entry point to all off-chain communication. 
We should either keep it as is (and adapt `/command` notation here) or change it across all LIPs to be a logical representation of resources and actions. 
I also think that we should stick to the main principles of LIP-1 [CommandRequestObject](https://lip.libra.org/lip-1/#commandrequestobject) and [CommandResponseObject](https://lip.libra.org/lip-1/#commandresponseobject) as it makes things easier for developers to have a one consistent protocol scheme/format.

andll: We specifically chose not to use command request/response style here to keep it simpler

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+}
+```
+
+| Field 	      | Type  | Required? 	| Description 	           |
+|-------	      |------ |-----------	|-------------	           |
+|libra_id         |str    | Y           | Libra ID of the user |
+|display_name     |str    | N           | Display name of the user. Usually a full name |
+|profile_picture  |str    | N           | Profile picture URL(see below) |
+
+If a user does not exist, VASP can choose to return an HTTP error, or the empty User Object, depending on it’s privacy choices.
```
LuZhang-Lou, line 220: Can we explicitly set a message for "not intended to share information", or "information unknown"? I think this helps clarify the results for sender, so they have better wordings to communicate to their users.

lips/lip-10.md
LuZhang-Lou: @bmaurer  can you explain the hot links and how FB proxies it?

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Profile Picture URI
+Profile picture URI is a URI that represents a profile picture of a user. Wallet app must support two possible URI schemas for profile picture.
+
+* **Https schema.** If URI starts with https:, wallet must attempt to fetch profile picture from the HTTP URI provided. When fetching this URI wallet app will not provide any authentication. Wallet app will use GET method to fetch the profile picture specified using https schema.
```
LuZhang-Lou, line 225: > When fetching this URI wallet app will not provide any authentication.

So basically it means VASPs want to expires URIs often.

lips/lip-10.md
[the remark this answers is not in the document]
danprinz: My question was regarding where in the request can one find the sending VASP address. It either be written in the request body, or as part of the path (like in LIP-1). 
If the receiver has the address at hand it should be easy to verify the request content. 

If LIP-1 is going to be changed to any other format (I would suggest a more RESTful one) we should keep this one as close and consistent as we can.

lips/lip-10.md
danprinz: That's right. I just wondered if OpenAPI spec would be more readable and common to use. Maybe we should adapt it in LIP-1 as well

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+* For both endpoints, all communications are signed via the same mechanism as for other off-chain communications, see the [Payloads section of lip-1](https://github.com/libra/lip/blob/master/lips/lip-1.mdx#payloads) for details. As such, all vasp-2-vasp communications are authenticated.
+
+* All Libra ID endpoints are nested under `pay.<version>` URI path:
+
+     `https://<hostname>:<port>/pay.<protocol_version>/<endpoint>`
+
+* All Libra ID endpoints use the same `base_url` and same `compliance_public_key` for authentication as other off-chain APIs.
+
+* All Libra ID endpoints are HTTP POST endpoints, they accept JSON-encoded requests and return JSON-encoded responses. Other encoding methods can be considered in the future, `Content-Type` and `Accept` HTTP headers can be used to define request/response formats.
```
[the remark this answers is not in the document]
andll: 1. It uses POST because GET requests don't have request body 👍

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+Furthermore, a user can have multiple Libra IDs within the same VASP. This does not affect the protocol, but VASPs can leverage this - for example to allow users to create multiple ‘accounts’ like checking, savings, etc.
+
+VASPs can also allow users to generate temporary / situational Libra IDs. This can be used in many scenarios such as:
+
+* Temporary Libra ID for some merchant payments, when user don’t want to share their ‘main’ Libra ID with the merchant
+* Publicly shared Libra IDs for the various campaigns, like donations, etc.
+
+## Libra ID domains
+Each Libra ID domain is associated with a VASP. The source of truth for this association is a blockchain - a special LibraIdDomains resource must be published into VASP for it to expose its LibraID domain.
```
kphfb, line 55:
```suggestion
Each Libra ID domain is associated with a single VASP. The source of truth for this association is the blockchain - a special LibraIdDomains resource must be published into VASP for it to expose its LibraID domain.
```

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+* Only special Treasury Compliance account (address `0xB1E55ED`) can manipulate LibraIdDomains resource:
+  * Only TC account can create and publish LibraIdDomains resource
+  * Only TC account can add, remove or update an LibraIdDomain within LibraIdDomains resource
+* In order to register a Libra Id domain, VASP needs to submit a request to Libra Association. LA will perform certain checks (out of scope of this document) before they will issue an on-chain transaction to register a Libra Id Domain. The checks that LA will perform will make sure that VASPs do not claim irrelevant Libra Id Domains. LA submission will also ensure that Libra Id Domains are unique.
+* It should also be possible (and probably going to be most common) for a new VASP to submit their desired Libra ID domain(s) along with the request to create a `ParentVASP` account.
+* Association may provide an open source application(*Indexer*) that will allow indexing available Libra IDs and will provide a convenient REST api for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
```
kphfb, line 75:
```suggestion
* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Libra IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
```

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+# Peer-2-peer payment protocol
```
kphfb, line 77: Can we make this something like "Utilization of LibraID during payments" and not tie it so specifically to p2p payments?  I kind of think we should tweak this whole section to make it just describe how libra ID works under the hood.  The internals of it should be the same for any flow that is used. I think even for merchant flows, the APIs should end up working in a very similar way.  I would expect a merchant flow to really just layer on top of the typical payment case.  So in a merchant flow:
* Enter your libraID on the merchant's page
* some magic happens
* app-switch to your wallet app
* your wallet pre-fills the txn details (inclusive of the receiver's libra ID)
* from this point, for a general merchant payment (and not an auth, recurring, etc.), the merchant wrapper ends and the flow described in this doc happens

I think that if we describe the base case here, it makes it easier to layer things on top of it (i.e. merchant) and helps with understanding without tying this to a specific p2p case only

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+One of the main use cases for Libra ID is transferring money from one user to another across different wallets.
+
+<img alt="Payment flow" src={useBaseUrl('img/libra-id-flow.png')} />
+
+## UI flow
+* Receiver shares their Libra ID with sender
```
kphfb, line 83:
```suggestion
* Receiver shares their Libra ID with sender (outside the scope of this document)
```

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+
+* **Info endpoint** for querying optional information about the receiver. VASPs can provide this endpoint to improve UX and security of payment, but they are not obligated to - VASPs may choose not to share any information about the user via this endpoint, including whether the user exists or not.
+  * *For example*: VASP can have notion of 'contact list' and only share information if the sender is in receiver's contacts
+* **Payment endpoint** to perform actual money transfer.
+
+Common considerations:
+
+* For both endpoints, all communications are signed via the same mechanism as for other off-chain communications, see the [Payloads section of lip-1](https://github.com/libra/lip/blob/master/lips/lip-1.mdx#payloads) for details. As such, all vasp-2-vasp communications are authenticated.
+
```
kphfb, line 100: I think a lot of this and down makes it seem like this is separate but similar to LIP-1, but to me it makes more sense to state it as an extension of LIP-1 that adds new functionality (basically a query API and a new command in the normal API).  Then you don't really need to re-state the things of LIP-1 which I think kind of add confusion since we only state some of the things

lips/lip-10.md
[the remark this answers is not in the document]
kphfb: Agree with this comment.  Let's match LIP-1 and instead of command add a new action (query?  get_libra_id?  etc.)

lips/lip-10.md
kphfb: I'd agree with @danprinz here.  If we are going to treat this as an action, then it should be using the same command structure as is used in LIP-1 and LIP-8.  That keeps us consistent and also means we don't need to re-state so much and make it seems sort of like LIP-1 but sort of its own thing

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+   "reference_id": "<Payment_Reference_ID>",
+   "receiver": "<User Object>"
+}
+```
+
+| Field 	      | Type       | Required? 	| Description 	           |
+|-------	      |------      |-----------	|-------------	           |
+|receiver_address |str         | Y            | On-chain account address for receiving the payment. This address must be encoded as a [bench 32 account identifier](https://github.com/libra/lip/blob/master/lips/lip-5.md#account-identifiers) **without** subaddress |
+|reference_id     |str         | Y            | Reference ID to be included into payment metadata |
+|receiver         |User Object | Y            | Information about the receiver |
```
kphfb, line 186: Make this a link to user object section

lips/lip-10.md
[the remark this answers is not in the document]
davidiw: Let's update here

lips/lip-10.md
davidiw: I don't follow, that compliance key is an ed25519 key for signing metadata that appears on chain. I don't know why we need a compliance key in this place that is strictly used for dual attestation. This is too overloaded and has no meaning in the rest of this document. Also if we add an IP for each address, what are we doing? Can we clarify the purpose of a wallet uri?

lips/lip-10.md
davidiw: Please update to LIP1 but why are we repeating this information?

lips/lip-10.md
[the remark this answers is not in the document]
davidiw: Usually a POST implies modification of data or a sh** ton of data, not sure if either of these apply.

lips/lip-10.md
davidiw: Yes and let's use 404.

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+* Receiver VASP may optionally choose how much information to share based on the sender Libra ID. If Receiver VASP decides to not share data, it should return an empty `User Object` with only `libra_id` field set
+* If user is not found, the VASP should return error code `400`  
+
+## Pay endpoint
+
+```
+POST https://<hostname>:<port>/pay.<protocol_version>/pay
+```
+
+**Pay endpoint** is used when a user is willing to make a money transfer.
```
davidiw, line 144: Why are we doing this off chain? It feels like I can embed some metadata in the transaction and receiving vasp  can use the lookup service... this feels a little unnatural.

lips/lip-10.md
davidiw: This is an open write API, this is dangerous. Folks will mess this up and have vulnerabilities.

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+* In order to register a Libra Id domain, VASP needs to submit a request to Libra Association. LA will perform certain checks (out of scope of this document) before they will issue an on-chain transaction to register a Libra Id Domain. The checks that LA will perform will make sure that VASPs do not claim irrelevant Libra Id Domains. LA submission will also ensure that Libra Id Domains are unique.
+* It should also be possible (and probably going to be most common) for a new VASP to submit their desired Libra ID domain(s) along with the request to create a `ParentVASP` account.
+* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Libra IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
+
+# Utilization of LibraID during payments
+One of the main use cases for Libra ID is transferring money from one user to another across different wallets.
+
+<img alt="Payment flow" src={useBaseUrl('img/libra-id-flow.png')} />
+
+## UI flow
```
davidiw, line 72: This is your story. The story lacks a lot of details that are specified below, such as the receiver side experience.

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+}
+```
+
+| Field 	      | Type       | Required? 	| Description 	           |
+|-------	      |------      |-----------	|-------------	           |
+|receiver_address |str         | Y            | On-chain account address for receiving the payment. This address must be encoded as a [bench 32 account identifier](https://github.com/libra/lip/blob/master/lips/lip-5.md#account-identifiers) **without** subaddress |
+|reference_id     |str         | Y            | Reference ID to be included into payment metadata |
+|receiver         |[User Object](https://github.com/libra/lip/blob/master/lips/lip-10.md#user-object) | Y            | Information about the receiver |
+
+* Pay endpoint is used to exchange details of the payment such as information about Sender and Receiver off the chain, and negotiate `Payment_Reference_ID`
```
davidiw, line 178: we should establish a clearer user story as an example before getting extremely technical about it. That is really important because it is this step that gains you your ability to distinguish a payment id. We should look into protocols like syn cookies for TCP that allow zero overhead negotiation of secrets.

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+
+* Once the sender learns Payment_Reference_ID, it will submit an on-chain transaction including the `Payment_Reference_ID` as a metadata. Receiver will use the `Payment_Reference_ID` in the transaction to link the on-chain transaction to the payment.
+
+* Sender will submit a transaction to the `receiver_address` on-chain account in the response. This account address can be either Parent VASP, or any of Child VASP accounts of the receiver VASP. This is needed to allow the receiver VASP to manage load between different on-chain accounts.
+
+* `Payment_Reference_ID` must be unique for each payment, it must not be reused in any way to avoid leaking privacy information on-chain.
```
davidiw, line 184: who enforce this? in order for this to be meaningful we would need this to be enforced by at least one VASP, they probably won't unless we have a certification process, or the blockchain.

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+* `Payment_Reference_ID` returned by this endpoint is not signed for KYC purposes. This means that payment above KYC threshold can not be submitted right away using the provided Payment_Reference_ID. Instead, sender VASP needs to proceed to KYC endpoint to negotiate the signed Reference ID.
+
+* KYC endpoint API will be updated to accept `Payment_Reference_ID` returned by the Libra ID endpoint instead of address/subaddress.
```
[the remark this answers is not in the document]
davidiw: agreed, we would want to state there is an additional KYC API that supports this... alternatively I would recommend that the payment_reference_id be subaddress

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+
+## User Object
+User Object defines reach information about the user.
+
+All fields, except for `libra_id`, in this object are optional - VASP can choose to share either of those fields or none.
+
+Additionally, all fields in this object are informational - **they must not be used for KYC and only intended for UI purposes**.
```
davidiw, line 195: rewrite are information, that is they are intended for UI purposes only and explicitly not establishing KYC.

lips/lip-10.md
davidiw: I would strike "depending on it's privacy choices"

If a user does not exist or has enabled contact-list mode, the VASP may choose to return either an HTTP 404 error or an empty User Object.

lips/lip-10.md
davidiw: I think this is an implementation detail that someone can easily screw up as long as the URL returns an image we are good to go.

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+## Profile Picture URI
+Profile picture URI is a URI that represents a profile picture of a user. Wallet app must support two possible URI schemas for profile picture.
+
+* **Https schema.** If URI starts with https:, wallet must attempt to fetch profile picture from the HTTP URI provided. When fetching this URI wallet app will not provide any authentication. Wallet app will use GET method to fetch the profile picture specified using https schema.
+  * Ex.: *https://novi.com/pictures/alice.png*
+* **Data schema.** If URI starts with data:, wallet must interpret the URI as a [Data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) that contains profile picture encoded.
+  * Ex.: *data:image/gif;base64,AAAAAAAA*
+* **Unencrypted http** schema must not be used to share profile pictures.
+
+Vasps can return a user profile picture in either format. HTTPS schema is more traditional, however using data schema allows to show profile pictures faster and avoid an extra HTTP request between VASPs.
```
davidiw, line 221: not required.

davidiw (changes requested):
Lots of comments.... overall the story does make sense, here are my major blockers before moving this to draft:

1) Clean up the writing and make it tighter -- most of my feedback is in this category -- in general, I might recommend that we avoid crappy editor frameworks like GitHub and consider a first draft in a shared editor but let's save that battle for another day :D
2) The user story should be near the top of this document and tie together the two properties provided: easy use and anonymity on chain
3) The document is not particularly searchable the profile_picture_uri is only contained once, you should reuse the exact format in the section that describes it so that idiots like me can easily search for it
4) diminutives / abbreviations should not be in technical documents examples: txn and app
5) Unify the URL scheme with Lip-1
6) Each portion of the Lip should be very clear -- what is the wallet_uri, why are we discussing the compliance endpoint and key, what is the specific use case of each endpoint
7) Having a writable endpoint is a no go for me, I won't block on this principle but there are really great techniques to support time valid identifiers that require a zero write API -- for example get_reference_id(user_id, txn_metadata) -> hash(secret_receiver, user_id, txn_metadata)
8) Would recommend considering how to make the payment reference id compatible with a subaddress otherwise we're going to have confusing APIs or alternatively define a scheme that makes it clear that one is a libra id and the other is a subaddress

Feel free to setup some time with me if that would help.

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+
+In addition to describing the format of the Libra ID, this LIP also describes the off-chain protocol for the peer-2-peer payment with Libra ID.
+
+# Motivation
+Libra ID provides a convenient way to identify a user's account within a VASP. There are two main use cases for Libra ID:
+* For peer-2-peer payments, Libra ID allows a user to provide a human-readable ‘identifier’ to the sender. Basically, instead of sending a [lip-5 address](https://github.com/libra/lip/blob/master/lips/lip-5.md) (that needs to be refreshed for each payment, ideally), the receiver can just share their Libra ID once. In this case Libra ID plays the role of an email address for payments. Benefits for this use case:
+  * Libra ID is a user-friendly, human-readable identifier, unlike lip-5 addresses
+  * The Libra ID resolution process can facilitate the exchange of rich information(e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
+  * Libra ID provides better privacy compared to using sub-addresses, because Libra ID is not shared on the chain and will always use a single-use identifier for the on-chain ID
```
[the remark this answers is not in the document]
andll: This is not always true.
Example is this - you and friends went to restaurant and you paid for everyone. Now you have a group chat in <messenger of your choice> and want to share your Libra wallet address so that they can pay you back.

There is no convenient way to share lip-5 address so that participants  can only send money, but not track payments from others. Of course in theory you can create subaddress for each, but this is not practical

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+
+Example: `andrey$novi`
+
+* **user-id**, identifies an account within a VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]*$`
+  * Maximum length: 64 characters
+  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
+  * Each VASP can create the user-id however they see fit - user selected, auto-generated, etc. and these are not required to be unique across different VASPs because the domain provides disambiguation across different VASPs. This user-id is a parallel to the user ID in an email address
+* **libra-id-domain** identifies the VASP:
```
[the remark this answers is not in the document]
andll: I think the plan is that it does not `uniquely` identify VASP, in a sense that VASP can have multiple libra id domains registered with it

lips/lip-10.md
```diff
@@ -0,0 +1,221 @@
...
+  * Libra ID is a user-friendly, human-readable identifier, unlike lip-5 addresses
+  * The Libra ID resolution process can facilitate the exchange of rich information(e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
+  * Libra ID provides better privacy compared to using sub-addresses, because Libra ID is not shared on the chain and will always use a single-use identifier for the on-chain ID
+  * From a user's perspective, Libra ID is a persistent identifier.  But from the perspective of the chain, LibraIDs do not exist and instead transactions are have a single-use ID to identify them. By contrast, from the viewpoint of both a user and the chain, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy, but appears on-chain and could thus be used to link payments if it isn't recycled upon each usage.
+
+# Libra ID format
```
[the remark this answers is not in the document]
andll: Why should format be capitalized? Should we capitalize every word in header?

andll:
>Having a writable endpoint is a no go for me, I won't block on this principle but there are really great techniques to support time valid identifiers that require a zero write API -- for example get_reference_id(user_id, txn_metadata) -> hash(secret_receiver, user_id, txn_metadata)

I am not sure I understand, how does generating reference id with libra id different comparing to generating reference ID during KYC? Are you saying it is bad in both cases, or there is something special about libra id, that does not make it work here, but it works for KYC?

>Would recommend considering how to make the payment reference id compatible with a subaddress otherwise we're going to have confusing APIs or alternatively define a scheme that makes it clear that one is a libra id and the other is a subaddress

There is a hope that with Libra ID we can actually deprecate subaddresses altogether, given their privacy implications.

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+
+* **user-id**, identifies an account within a VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]*$`
+  * Maximum length: 64 characters
+  * Not published on the chain, there is no way for an external entity to list all the users within VASP.
+* **libra-id-domain** identifies the VASP:
+  * Case insensitive
+  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]+$`
+  * Maximum length: 63 characters
```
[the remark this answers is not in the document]
andll: I think usually you want to have length limits on any data structures. For example, some VASPs are probably going to store Libra IDs in SQL database, and they might want to use some column type that have length limit

lips/lip-10.md
```diff
@@ -0,0 +1,231 @@
...
+Each Libra ID domain is associated with a VASP. The source of truth for this association is a blockchain - a special LibraIdDomains resource must be published into VASP for it to expose its LibraID domain.
+
+```
+resource struct LibraIdDomains {
+    domains: vector<LibraIdDomain>,
+}
+
+struct LibraIdDomain {
+    domain: vector<u8>,     // Utf-8 encoding
+    wallet_uri: vector<u8>, // ASCII encoding (use url-escape for non-ASCII characters)
```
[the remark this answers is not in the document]
andll: This is related for merchant, since right now we don't have LIP for merchant flow, I am just going to remove it

davidiw (approved):
welcome to the club

Let's continue to tighten this up in ensuing copies:
* clearer story around building an indexer / move events
* how this works with lip 1
* general massaging with flow as we go

PR #84, opened 2020-11-17
[LIP-1] Simplify protocol, specify explicit protocol flows & provide test vectors.

gdanezis:
This is a major change to the LIP-1 protocol following feedback from implementors. The aim of the update is to simplify the flows allowed, to help conforming independent implementations. Major changes include:

* Transport: Simplified URL, detailed JWS configurations, and test vectors.
* Sequencing: removal of read/write sets and object versions.
* Payment Commands: strict state-machine definition of the Payment protocol; in which Sender/Receiver take turns to update object or abort. Relaxation of formatting requirements for some fields which created confusion.


lips/lip-1.mdx
```diff
@@ -11,92 +11,67 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
 **Channel**: The communication path between a pair of entities who execute Commands and track the evolution of a set of Shared Objects.
 
-**Shared Object**: All Objects contained within a Command - for example `PaymentObject`s, are considered as "Shared Objects" - meaning that either VASP may create a new Command to modify the Object, and will do so during the typical life-cycle of an Object - an example being the addition of required data within the KycDataObject from both VASPs to a payment Object. Both VASPs in a Channel can asynchronously attempt to initiate and execute Commands on Shared Objects.
+**Shared Object**: All Objects contained within a Command - for example `PaymentObject`s, are considered as "Shared Objects" - meaning that either VASP may create a new Command to modify the Object, and will do so during the typical life-cycle of an Object - an example being the addition of required data within the `KycDataObject` from both VASPs to a payment Object. Only one VASPs in a Channel can may at any time attempt to initiate and execute a Commands on Shared Objects, defined by the state of the Object.
```
kphfb, line 45:
```suggestion
**Shared Object**: All Objects contained within a Command - for example `PaymentObject`s, are considered as "Shared Objects" - meaning that either VASP may create a new Command to modify the Object, and will do so during the typical life-cycle of an Object - an example being the addition of required data within the `KycDataObject` from both VASPs to a payment Object. Only one VASP in a Channel may at any time attempt to initiate and execute a Command on Shared Objects, defined by the state of the Object.
```

lips/lip-1.mdx
```diff
@@ -11,92 +11,67 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+The exposed endpoint receives JWS signed `CommandRequestObject`s in the POST body, and responds with a JWS signed `CommandResponseObject`s in the HTTP response (See [Request/Response Payload](#requestresponse-payload) for more details. Single Command requests-responses are supported (HTTP1.0) but also pipelined request-responses are supported (HTTP1.1). The content type of the HTTP request/response is ignored. The version for the Off-Chain Protocol is the string `0`.  All structures transmitted, nested within `CommandRequestObject` and `CommandResponseObject` are valid JSON serialized Objects and can be parsed and serialized using standard JSON libraries.
```
xli, line 69: `0` -> `v1`

lips/lip-1.mdx
```diff
@@ -498,19 +452,106 @@ Valid values are:
 * `needs_kyc_data` - KYC data about the subaddresses is required by this actor.
 * `needs_recipient_signature` - Can only be associated with the sender actor.  Means that the sender still requires that the recipient VASP provide the signature so that the transaction can be put on-chain.
```
xli, line 453: we said remove needs_recipient_signature, right?

lips/lip-1.mdx
```diff
@@ -498,19 +452,106 @@ Valid values are:
...
+
+* SINIT: (`need_kyc_data`, `none`, `*`, `*`)
+* RSEND: (`need_kyc_data`, `ready_for_settlement`, `*`, `*`)
+* RABORT: (`need_kyc_data`, `abort`)
+* SABORT: (`abort`, `ready_for_settlement`, `*`, `*`)
+* READY: (`ready_for_settlement`, `ready_for_settlement`, `*`, `*`)
+
+Soft-match disambiguation states:
+
+* RSOFT (`need_kyc`, `soft_match`, `_`, `_`)
```
xli, line 475: need_kyc => need_kyc_data

lips/lip-1.mdx
```diff
@@ -498,19 +452,106 @@ Valid values are:
...
+* SSOFTSEND (`need_kyc`, `soft_match`, `Sx`, `_`)
```
xli, line 476: need_kyc => need_kyc_data

lips/lip-1.mdx
```diff
@@ -498,19 +452,106 @@ Valid values are:
...
+* SSOFT (`soft_match`, `ready_for_settlement`, `*`, `_`)
+* RSOFTSEND (`soft_match`, `ready_for_settlement`, `*`, `Rx`)
```
xli, line 478: is receiver allowed to abort when sender send soft match?
 (soft_match, abort)

similarly, when receiver send soft_match, could sender abort?

lips/lip-1.mdx
```diff
@@ -107,53 +77,24 @@ Assume two VASPs A and B.  The basic protocol interaction consists of:
...
-If the result is 0, the lexicographically lower parent address is used as the server side.
-If the result is 1, the lexicographically higher parent address is used as the server side. Lexicographic ordering determines which binary address is higher by comparing byte positions one by one, and returning the address with the first higher byte.
+Given the state of an object there is an unambigious way to define who the **next writer** to the object should be (unless the object is in a terminal state and no more commands should be issued on it). Only this VASP is allowed to issue a command to update this Object, and a command from the other VASP will be rejected as a command failure. This prevents concurrent attempts to update Objects that could lead to inconsistencies.
```
xli, line 88: this is not very clear about what it means of "unambigious way to define who the next writer", 
if I understand correctly, it means the command object (like PaymentObject) design should ensure the unambigious definition of next writer.
may add a ref link to below "PaymentObject Protocol Command Sequence" for example

lips/lip-1.mdx
```diff
@@ -493,27 +425,188 @@ Represents a national ID.
...
+The payment can be executed on-chain by the sender (or settled in any other way).
+
+### RSEND -> SABORT (Step 9)
+
+The Sender issues `PaymentCommand` to update an existing payment.
+
+The sender VASP requires more time, has a soft match for the receiver or cannot proceed with the payment. It issues an abort command:
+
+* `Sender.Status.Status`: `abort`
+* `Sender.Status.abort_code`: one of `pending`, `soft-match`, or `rejected`.
```
xli, line 514: why "soft-match" is abort code? what is difference comparing with sender send status = soft-match?

lips/lip-1.mdx
```diff
@@ -493,27 +425,188 @@ Represents a national ID.
...
+
+* `Sender.Status.Status`: `ready_for_settlement`
+
+The payment can be executed on-chain by the sender (or settled in any other way).
+
+### RSEND -> SABORT (Step 9)
+
+The Sender issues `PaymentCommand` to update an existing payment.
+
+The sender VASP requires more time, has a soft match for the receiver or cannot proceed with the payment. It issues an abort command:
```
xli, line 511: when it has soft match for the receiver, why abort?

lips/lip-1.mdx
```diff
@@ -493,27 +425,188 @@ Represents a national ID.
 ```
 
 ### StatusEnum
-Valid values are:
+
+Valid values are the unicode strings:
 * `none` - No status is yet set from this actor.
 * `needs_kyc_data` - KYC data about the subaddresses is required by this actor.
-* `needs_recipient_signature` - Can only be associated with the sender actor.  Means that the sender still requires that the recipient VASP provide the signature so that the transaction can be put on-chain.
 * `ready_for_settlement` - Transaction is ready for settlement according to this actor (i.e. the required signatures/KYC data have been provided)
```
kphfb, line 432: Do we still want this state?  Looking at the diagram below, not sure it actually means anything anymore

dahliamalkhi (changes requested):
This refactor seem to help a lot with clarity. I wonder whether we want to do it in LIP-1, or LIP-1.1, or a new LIP. 

Also, please address @xli 's questions.

lips/lip-1.mdx
```diff
@@ -11,92 +11,62 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-### Concurrency Control
-The Off-Chain Protocol manages concurrent requests on Objects and ensures that Object states will not become out of sync.
+**Version**: The state of an Object at a specific point in time. Upon creation an Object is assigned version 0. Every creation or mutation increases the Version of the Object which was created/mutated by one.  VASPs operate upon the latest Version of Shared Objects.  Once a Version is mutated, it is no longer a valid Version upon which a new Command may depend - the newly updated value of the Version is the latest Version for that Object and must be acted upon.
 
-### Eventual Consistency
-The Off-Chain Protocol ensures eventual consistency of Shared Objects.
-
-# Basic Off-Chain Building Blocks
+**reference_id**: Every Object type contains a `reference_id` field which is a unique reference ID for the Object. The reference ID is always specified by the payment initiator VASP (the VASP which originally created the Object). This value should be unique. We recommend the hexadecimal representation of a fresh >16 bytes random byte-string as a way of creating such values. The field is Unicode utf-8 encoded.
```
LuZhang-Lou, line 49: I want to bring up this question again: 
What are the benefits of allowing utf-8 charsets v.s plain and simple alphanumeric? 
Using utf-8 charsets complicates databases (e.g. MySQL) operations and needs extra caution to deal with regard SQL injection.

andll: Do you guys think it makes sense to add regex to reference_id (and other fields)? I think allowing _any_ character for reference id is certainly a bad thing

xli: could we consider UUID only?

lips/lip-1.mdx
gdanezis: Done!

lips/lip-1.mdx
gdanezis: Its gone :)

lips/lip-1.mdx
xli: if someone put in sequence: 1, 2, 3 from their db index, is it ok? when 2 vasps do this, it's a problem right?

andll: We should probably say that unique identifier of a payment is (sender, receiver, ref_id), so different vasp don't need to worry about clash with other vasps

lips/lip-1.mdx
```diff
@@ -534,7 +622,7 @@ metadata: libra.Metadata = libra.Metadata__TravelRuleMetadata(
 )
 
 # on-chain account address in hexadecimal (https://github.com/libra/lip/blob/master/lips/lip-5.md#terminology)
-sender_libra_address_hex = "793251de1a61e9b4d1a17d13aa015e45  
+sender_libra_address_hex = "793251de1a61e9b4d1a17d13aa015e45
```
LuZhang-Lou, line 625:
```suggestion
sender_libra_address_hex = "793251de1a61e9b4d1a17d13aa015e45"
```

lips/lip-1.mdx
```diff
@@ -11,149 +11,87 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-
-* `protocol_version` is initially 0 for this iteration of the off-chain APIs.
-* `LocalVASPAddress` and `RemoteVASPAddress` are the bech32 VASP addresses of the respective VASPs - with the subaddress set to 0's as per [LIP 5](https://lip.libra.org/lip-5/).
-    * An example URL: http://127.0.0.1:7091/v1/lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m/lbr1pgfpyysjzgfpyysjzgfpyysjzggqqqqqqqqqqqqqxjt54z/command/
-
+Each VASP exposes an HTTPS POST end point at `https://<hostname>:<port>/<protocol_version>/command`. The `protocol_version` is initially `v1` for this iteration of the off-chain APIs.
 
 ### HTTP Headers
-All HTTP requests and responses must contain a header `X-Request-ID` with a unique ID for the request, used for tracking requests and debugging. Responses must have the same string in the `X-Request-ID` header value as the requests they correspond to.
+All HTTP requests and responses must contain:
```
LuZhang-Lou, line 61: let's be explicit here x-request-id and others are case-insensitive.

lips/lip-1.mdx
```diff
@@ -11,149 +11,89 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-* `protocol_version` is initially 0 for this iteration of the off-chain APIs.
-* `LocalVASPAddress` and `RemoteVASPAddress` are the bech32 VASP addresses of the respective VASPs - with the subaddress set to 0's as per [LIP 5](https://lip.libra.org/lip-5/).
-    * An example URL: http://127.0.0.1:7091/v1/lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m/lbr1pgfpyysjzgfpyysjzgfpyysjzggqqqqqqqqqqqqqxjt54z/command/
+### HTTP Headers
+All HTTP requests and responses must contain:
 
+* A header `X-REQUEST-ID` with a unique ID for the request, used for tracking requests and debugging. Responses must have the same string in the `X-REQUEST-ID` header value as the requests they correspond to.
+* A header `X-VERIFICATION-KEY-ADRRESS` with the LIP-5 Libra address of the VASP signing the HTTP reqquest or HTTP response. This header is used to retrieve the VASP verification key used to authenticate the request or response. For HTTP requests, the header is set to the VASP parent address initiating the request. For HTTP response, the header is set to the parent VASP address of the VASP responding.
```
xli, line 64: may use VASP parent / DESIGNATED DEALER

LuZhang-Lou, line 64: I changed my view with this `X-VERIFICATION-KEY-ADRRESS`. I think what should be here is just the acting VASP's address. Two reasons:

1. This helps with the receiving VASP to identity who it is talking to more quickly. It's possible that a VASP only wants to talk to VASP A, but not VASP B, even though they are all children of VASP C. Using `verification-key-address` makes VASPs need to dig into the payloads (and doing expensive crypto ops) to see the exact VASP address. This is not an imaginary scenario, Novi deals with this today.

2. A sound verification should check the address in the header matches the one in PaymentActor (which I think current implementation misses too?). Now this may be subjective, but checking `the one in header == the one in PaymentActor` is more intuitive than checking `the one in header parents the one in Payment Actor`.

The only advantage of having `verification-key-address` in header I think is the convenience of getting the compliance key. But stepping back a bit, wouldn't it be more convenient to directly attach the public key in header?

xli: use address will make sure it is a valid onchain vasp / dd account, given onchain account creation process. attach public key will make it open to everyone.

gdanezis: My thoughts on this:

* I like the idea of having here the acting VASP, rather than the parent VASP. A correct implementation should check that the acting VASP has a payload signed by the parent VASP, and this indirection forces the check to retrieve the key.
* Due to the previous point, I would prefer to not attach the verification key. If we do this invited implementors to not check the binding between VASP - Parent VASP - verification key as they should.

PS I think the current offchain implementation does not deal with parent VASPs, and instead assumes it is given a direct tuple of (acting VASP, parent key). Then then checks do happen in the check-payment logic I believe.

lips/lip-1.mdx
gdanezis: Ok, I added: Both 

> `X-REQUEST-ID` and `X-VERIFICATION-KEY-ADRRESS` are case insensitive.

lips/lip-1.mdx
```diff
@@ -11,149 +11,87 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-In the future, the Off-Chain Protocol may be further extended to include functionality such as batching of transactions and additional payments functionality.
+## Overview
 
----
-# Specification
----
-
-# Off-Chain Protocol Design Principles
+Two VASPs participate in the off-chain protocol. They communicate through HTTP request and responses protected with TLS, and exchange messages that are signed to ensure authenticity and integrity. The goal of the off-chain protocol is for a pair of VASPs to jointly create a `PaymentObject` containing a reference-id. When a `PaymentObject` is **completed**, both VASPs have identical copies of it, and both and can map its reference-id to the object. A completed `PaymentObject` can be submitted by either VASP to the Blockchain to settle.
```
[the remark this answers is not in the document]
gdanezis: Thanks fixed!

lips/lip-1.mdx
```diff
@@ -492,26 +418,188 @@ Represents a national ID.
...
+
+**Protocol Flow Illustration & VASP logic**. Below is a high level illustration of the commands and internal processes a sender and receiver VASPs to define a payment, and update it until it reaches a final state. Light blue blocks represent successful PaymentCommand requests and responses that create or update the PaymentObject. Each is labelled from 1-9, and we will reference these labels when discussing each step in details below.
+
+<img alt="Command Exchange" src={useBaseUrl('img/LIP1-PaymentBusinessFlow.png')} />
+
+## Steps of the protocol: The KYC Exchange Flow
+
+The basic KYC exchange flow starts with the Sender including KYC information and requesting KYC information by the receiver. In the straight-forward case the Receiver is satisfied with this information and responds with a command providing the receiver KYC information, and recipient signature. The Sender can then finalize the payment and also settle it on-chain (status sequence SINIT -> RSEND -> READY). At each step Sender or Receiver may also abort the payment.
```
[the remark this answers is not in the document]
gdanezis: Nice suggestion: I used your suggested language.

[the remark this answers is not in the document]
gdanezis: I tried to include some more insight, but the full description follows after this short overview.

lips/lip-1.mdx
```diff
@@ -107,53 +77,24 @@ Assume two VASPs A and B.  The basic protocol interaction consists of:
...
 
 # Network Error Handling
 
-In the case of network failure, the sending party for this Command is required to re-send the Command until it gets a response from the counterparty VASP.  An exponential backoff is suggested for Command re-sends.  This retransmission strategy allows the Off-Chain Protocol to achieve [eventual consistency](#eventual-consistency) of the Shared Objects.  Retransmission must be done regardless of the role of the VASP (protocol client or protocol server).
+In the case of network failure, the sending party for this Command is required to re-send the Command until it gets a response from the counterparty VASP.  An exponential backoff is suggested for Command re-sends.  This retransmission strategy allows the Off-Chain Protocol to achieve [eventual consistency](#eventual-consistency) of the Shared Objects.
```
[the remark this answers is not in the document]
gdanezis: This does create a complication: the initiating VASP may not know if the other VASP has received and processed the message. So it cannot assume that because it did not get a response yet, it may never get a response?

lips/lip-1.mdx
```diff
@@ -11,92 +11,62 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+The exposed endpoint receives JWS signed `CommandRequestObject`s in the POST body, and responds with a JWS signed `CommandResponseObject`s in the HTTP response (See [Request/Response Payload](#requestresponse-payload) for more details. Single Command requests-responses are supported (HTTP1.0) but also pipelined request-responses are supported (HTTP1.1). The content type of the HTTP request/response is ignored. The version for the Off-Chain Protocol is the string `v1`.  All structures transmitted, nested within `CommandRequestObject` and `CommandResponseObject` are valid JSON serialized Objects and can be parsed and serialized using standard JSON libraries.
```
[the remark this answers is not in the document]
gdanezis: Yeah, that is odd. Removed since we say it before along with URL format.

lips/lip-1.mdx
gdanezis: Good point @andll !

lips/lip-1.mdx
```diff
@@ -319,32 +251,25 @@ For a travel rule data exchange, the [command_type](#commandrequestobject) field
...
-    "_writes": {
-            "lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m_d4115900aad65d8141236b464f1315f5": "12ce83f66d180d6e08b6c00fdbbf085a"
-    },
     "payment": {
         PaymentObject(),
     }
 }
 ```
 
+The contents of `payment` in a `PaymentCommand` either create or update a `PaymentObject`. The field `reference_id` must always be included in `payment` to determine which object is created or updated. In case of update there is an existing Object with the given `reference_id`, and the contents of the `payment` field update or add to the fields that already exist -- validation happens on the resulting, updated, object. This enables transfering only the updated fields within a `PaymentCommand`, rather than retransmitting the full state of the `PaymentObject` (which is also valid).
```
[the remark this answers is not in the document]
gdanezis: Simpler is better, so I removed this section.

lips/lip-1.mdx
```diff
@@ -371,7 +296,8 @@ A `PaymentActorObject` represents a participant in a payment - either sender or
...
 | kyc_data | [KycDataObject](#kycdataobject) | N | The KYC data for this account. This field is optional but immutable once it is set. |
 | status | [StatusObject](#statusobject) | Y | Status of the payment from the perspective of this actor. This field can only be set by the respective sender/receiver VASP and represents the status on the sender/receiver VASP side. This field is mandatory by this respective actor (either sender or receiver side) and mutable. |
-| metadata | list of str | Y | Can be specified by the respective VASP to hold metadata that the sender/receiver VASP wishes to associate with this payment. This is a mandatory field but can be set to an empty list (i.e. `[]`). New string-typed entries can be appended at the end of the list, but not deleted.
+| metadata | list of str | N | Can be specified by the respective VASP to hold metadata that the sender/receiver VASP wishes to associate with this payment. It may be set to an empty list (i.e. `[]`). New string-typed entries can be appended at the end of the list, but not deleted.
```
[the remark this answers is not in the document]
gdanezis: Removed it.

lips/lip-1.mdx
```diff
@@ -492,26 +418,188 @@ Represents a national ID.
...
+# PaymentObject Protocol Command Sequence
+
+**PaymentObject State.** The state of a `PaymentObject` is used to determine which VASP is expected to issue a command next, and what information is expected to be included in the command to progress the payment. The state is determined by the tuple of the status of the Sender and Receiver Actors and two flags `Sx` and `Rx`. The exact fields in the payment object for Sender and Receiver actor status are `sender.status.status` and `receiver.status.status`. The flag `Sx` and `Rx` are set when the Sender or the Receiver has included `additional_kyc_data` in the actor field (`Sender.kyc_data.additional_kyc_data` and `Receiver.kyc_data.additional_kyc_data` respectively).
+
+The states (Sender Status, Receiver Status, Sx Flag, Rx Flag) are:
+
+Basic KYC exchange flow
+
+* SINIT: (`need_kyc_data`, `none`, `*`, `*`)
+* RSEND: (`need_kyc_data`, `ready_for_settlement`, `*`, `*`)
```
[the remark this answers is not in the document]
gdanezis: "SETTED" has a meaning in payments and finance, and is likely to confuse. This is why I used neutral terms like SEND and READY here.

lips/lip-1.mdx
```diff
@@ -262,16 +262,14 @@ For a travel rule data exchange, the [command_type](#commandrequestobject) field
...
 
 | Field 	    | Type 	| Required? 	| Description 	|
 |-------	    |------	|-----------	|-------------	|
 | sender/receiver | [`PaymentActorObject`](#paymentactorobject) | Required for payment creation | Information about the sender/receiver in this payment |
-| reference_id | str | Y | Unique reference ID of this payment on the payment initiator VASP (the VASP which originally created this payment Object). This value should be unique. This field is mandatory on payment creation and immutable after that. The field contains a subset of characters in the set [`A-Za-z0-9_-`] and has a maximum length of 64 characters. |
+| reference_id | str | Y | Unique reference ID of this payment on the payment initiator VASP (the VASP which originally created this payment Object). This value should be unique. This field is mandatory on payment creation and immutable after that. The field contains a subset of characters in the set [`A-Za-z0-9_-`] and has a maximum length of 64 characters. Should be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122). |
```
LuZhang-Lou, line 272: Remove
> The field contains a subset of characters in the set [`A-Za-z0-9_-`] and has a maximum length of 64 characters.

?

lips/lip-1.mdx
```diff
@@ -11,149 +11,89 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-Upon receipt of a Command that has already been processed (resulting in a success response or a Command error), the receiving side must reply with the same response as was previously issued. Requests that resulted in protocol errors (such as `wait`) may result in different responses. For example if VASP A sends a Requests to VASP B, and VASP B responds with a `wait` protocol error, later VASP A may re-send the same request, and VASP B may respond with a `success` or a different error.
+Upon receipt of a Command that has already been processed (resulting in a success response or a Command error), the receiving side must reply with the same response as was previously issued (successful commands, or commands that fail with a command error, are _idempotent_). Requests that resulted in protocol errors (such as `wait`) may result in different responses. For example if VASP A sends a Requests to VASP B, and VASP B responds with a `wait` protocol error, later VASP A may re-send the same request, and VASP B may respond with a `success` or a different error.
```
kphfb, line 93: Do we still have the wait error anymore?  It kind of seems like we wouldn't need it now?

lips/lip-1.mdx
```diff
@@ -183,7 +123,7 @@ All requests between VASPs are structured as `CommandRequestObject`s.
 | _ObjectType  | str    | Y | Fixed value: `CommandRequestObject`|
 | command_type | str    | Y | A string representing the type of Command contained in the request. |
 | command      | Command object | Y | The Command to sequence. |
-| cid | str    | Y     | A unique identifier for the Command. |
+| cid | str    | Y     | A unique identifier for the Command. Should be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122). |
```
kphfb, line 126: I can't recall why we need this anymore.  Isn't the header covering this now?

lips/lip-1.mdx
```diff
@@ -203,7 +143,7 @@ All responses to a CommandRequestObject are in the form of a CommandResponseObje
 | _ObjectType    | str      | Y             | The fixed string `CommandResponseObject`. |
 | status         | str      | Y             | Either `success` or `failure`. |
 | error          | List of [OffChainErrorObject](#offchainerrorobject) | N | Details on errors when status == "failure" |
-| cid            | str      | N | The Command identifier to which this is a response. |
+| cid            | str      | N | The Command identifier to which this is a response. Should be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122). |
```
kphfb, line 146: Same comment here

kphfb: And if we do end up not needing these, we need to remove them from all examples as well

lips/lip-1.mdx
xli: (sender, receiver, ref_id) as unique identifier won't work if you reference this in another payment object, it will be very confusing to figure out what is sender and receiver  of the refered id

lips/lip-1.mdx
xli: I think header x-request-id is per request, but the cid can be same if the command was processed with error and client side retried the command with a new http request

lips/lip-1.mdx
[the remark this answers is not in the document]
LuZhang-Lou: fair enough

lips/lip-1.mdx
kphfb: Discussed offline and cid is largely for debugging purposes right now.  It allows replayed requests with different request IDs to be seen as a single command and traced against eachtoehr

lips/lip-1.mdx
```diff
@@ -11,149 +11,122 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* Only allow one actor to send command for mutating object.
+* Removed Command `_writes` and `_reads` with object versioning.
+* Changed reference_id to use 128 bit UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122).
+* additional_kyc_data field is moved from KycDataObject to PaymentActorObject.
+* Protocol URL is changed, removed sender/receiver's address in the URL path, added X-VERIFICATION-KEY-ADRRESS HTTP header for looking up JWS verification key.
+* Removed payment actor StatusEnum: needs_recipient_signature and pending_review.
+* Required to set sender's kyc_data for initial PaymentCommand.
+* Required to set receiver's kyc_data and recipient signature when receiver actor's status is set to ready_for_settlement.
+* Updated corresponding example code.
```
kphfb:
```suggestion
* Simplifications to state management:
  * Allow only one actor to send command for mutating object.
  * Removed Command `_writes` and `_reads` with object versioning - no longer needed when only a single actor can mutate an object in each object state
  * Required to set sender's kyc_data for initial PaymentCommand.
  * Required to set receiver's kyc_data and recipient signature when receiver actor's status is set to ready_for_settlement.
  * Removed payment actor StatusEnum: needs_recipient_signature and pending_review - no longer needed with updated state flow.
* Changed reference_id to use 128 bit UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122).
* additional_kyc_data field is moved from KycDataObject to PaymentActorObject. Allows KycDataObject to be set once and never mutated.
* Protocol URL is changed, removed sender/receiver's address in the URL path, added X-VERIFICATION-KEY-ADRRESS HTTP header for looking up JWS verification key.
* Updated corresponding example code.
```

lips/lip-1.mdx
```diff
@@ -11,149 +11,122 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-To ensure a consistent view of Object states, every Object is versioned. Mutations may only occur upon the latest Version of an Object.
+**Command**: An instruction sent over a Channel to mutate/create one or more Objects. In the case of a mutation, this Command will depend upon the current value of one of more existing Shared Objects.
 
-### Concurrency Control
-The Off-Chain Protocol manages concurrent requests on Objects and ensures that Object states will not become out of sync.
+**Channel**: The communication path between a pair of entities who execute Commands and track the evolution of a set of Shared Objects.
 
-### Eventual Consistency
-The Off-Chain Protocol ensures eventual consistency of Shared Objects.
+**Shared Object**: All Objects contained within a Command are logically shared, meaning that each VASP has a copy of the Object and may create a new Command to modify it. For example, during the lifecycle of a `KycDataObject`, both VASPs will add information to it. The protocol does not support simultaneous updates to Objects. Rather, VASPs must take turns updating the Object.
```
kphfb, line 69:
```suggestion
**Shared Object**: All Objects contained within a Command are logically shared, meaning that each VASP has a copy of the Object and may create a new Command to modify it. For example, during the lifecycle of a `KycDataObject`, both VASPs will add information to it. The protocol does not support simultaneous updates to Objects. Rather, VASPs must take turns updating the Object - with turns specified as per the [command sequence](#paymentobject-protocol-command-sequence).
```

lips/lip-1.mdx
```diff
@@ -11,149 +11,122 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-# Basic Off-Chain Building Blocks
+**Reference ID**: Every Object type contains a `reference_id` field which is a unique reference ID for the Object. The reference ID is always specified by the payment initiator VASP (the VASP which originally created the Object). This value should be global unique. We recommend use a 128 bits long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122).
```
kphfb, line 71:
```suggestion
**Reference ID**: Every Object type contains a `reference_id` field which is a unique reference ID for the Object. The reference ID is always specified by the payment initiator VASP (the VASP which originally created the Object). This value should be globally unique. It is recommended to use a 128 bit long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122).
```

lips/lip-1.mdx
```diff
@@ -11,149 +11,122 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+The exposed endpoint receives JWS signed `CommandRequestObject`s in the POST body, and responds with a JWS signed `CommandResponseObject`s in the HTTP response (See [Request/Response Payload](#requestresponse-payload) for more details. Single Command requests-responses are supported (HTTP1.0) but also pipelined request-responses are supported (HTTP1.1). The content type of the HTTP request/response is ignored.  All structures transmitted, nested within `CommandRequestObject` and `CommandResponseObject` are valid JSON serialized Objects and can be parsed and serialized using standard JSON libraries.
```
kphfb, line 94:
```suggestion
The exposed endpoint receives JWS signed `CommandRequestObject`s in the POST body, and responds with a JWS signed `CommandResponseObject`s in the HTTP response (See [Request/Response Payload](#requestresponse-payload) for more details). Single Command requests-responses are supported (HTTP1.0) but also pipelined request-responses are supported (HTTP1.1). The content type of the HTTP request/response is ignored.  All structures transmitted, nested within `CommandRequestObject` and `CommandResponseObject` are valid JSON serialized Objects and can be parsed and serialized using standard JSON libraries.
```

lips/lip-1.mdx
```diff
@@ -11,149 +11,122 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+In the case of network failure, the sending party for this Command is required to re-send the Command until it gets a response from the counterparty VASP. An exponential backoff is suggested for Command re-sends.
 
-Upon receipt of a Command that has already been processed (resulting in a success response or a Command error), the receiving side must reply with the same response as was previously issued. Requests that resulted in protocol errors (such as `wait`) may result in different responses. For example if VASP A sends a Requests to VASP B, and VASP B responds with a `wait` protocol error, later VASP A may re-send the same request, and VASP B may respond with a `success` or a different error.
+Upon receipt of a Command that has already been processed (resulting in a success response or a Command error), the receiving side must reply with the same response as was previously issued (successful commands, or commands that fail with a command error, are idempotent). Requests that resulted in protocol errors may result in different responses.
+For example:
+
+* VASP A sends a request to VASP B failed with an unknown network error. VASP A retried the send request to VASP B.
```
kphfb, line 122:
```suggestion
* VASP A sends a request to VASP B which failed with an unknown network error. VASP A retried the send request to VASP B.
```

lips/lip-1.mdx
```diff
@@ -11,149 +11,122 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* VASP B received both requests, and processed them in parallel. However the first request that VASP A considered as failure acquired the object / data lock by reference id.
+* VASP B processed first request successfully and replied a success response, but VASP A didn’t receive it.
+* VASP B respond a protocol error for the second request as it can’t acquire the object / data lock by reference id.
```
kphfb, line 125:
```suggestion
* VASP B responded with a protocol error for the second request as it can’t acquire the object / data lock by reference id.
```

lips/lip-1.mdx
```diff
@@ -202,20 +175,20 @@ All responses to a CommandRequestObject are in the form of a CommandResponseObje
 |-------	     |------	|-----------	|-------------	|
 | _ObjectType    | str      | Y             | The fixed string `CommandResponseObject`. |
 | status         | str      | Y             | Either `success` or `failure`. |
-| error          | List of [OffChainErrorObject](#offchainerrorobject) | N | Details on errors when status == "failure" |
-| cid            | str      | N | The Command identifier to which this is a response. |
+| error          | OffChainErrorObject(#offchainerrorobject) | N | Details of the error when status == "failure" |
```
kphfb, line 178:
```suggestion
| error          | [OffChainErrorObject](#offchainerrorobject) | N | Details of the error when status == "failure" |
```

kphfb: You need the brackets still because that's the markdown for a link

lips/lip-1.mdx
```diff
@@ -344,7 +302,7 @@ The structure in this Object can be a full payment or just the fields of an exis
 | Field 	    | Type 	| Required? 	| Description 	|
 |-------	    |------	|-----------	|-------------	|
 | sender/receiver | [`PaymentActorObject`](#paymentactorobject) | Required for payment creation | Information about the sender/receiver in this payment |
-| reference_id | str | Y | Unique reference ID of this payment on the payment initiator VASP (the VASP which originally created this payment Object). This value should be unique, and formatted as “<creator_vasp_onchain_address_bech32>_<unique_id>”.  For example, ”lbr1x23456abcd_seqABCD“. This field is mandatory on payment creation and immutable after that. "unique_id" should be 64 or fewer characters. |
+| reference_id | str | Y | Unique reference ID of this payment on the payment initiator VASP (the VASP which originally created this payment Object). This value should be global unique. This field is mandatory on payment creation and immutable after that. We recommend use a 128 bits long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122). |
```
kphfb, line 305:
```suggestion
| reference_id | str | Y | Unique reference ID of this payment on the payment initiator VASP (the VASP which originally created this payment Object). This value should be globally unique. This field is mandatory on payment creation and immutable after that. We recommend use a 128 bits long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122). |
```

lips/lip-1.mdx
```diff
@@ -492,84 +451,234 @@ Represents a national ID.
...
+* `soft_match` - Actor's KYC data resulted in a soft-match.
+
+# PaymentObject Protocol Command Sequence
+
+**PaymentObject State.** The state of a `PaymentObject` is used to determine which VASP is expected to issue a command next, and what information is expected to be included in the command to progress the payment. The state is determined by the tuple of the status of the Sender and Receiver Actors and two flags `Sx` and `Rx`. The exact fields in the payment object for Sender and Receiver actor status are `sender.status.status` and `receiver.status.status`. The flag `Sx` and `Rx` are set when the Sender or the Receiver has included `additional_kyc_data` in the actor field (`Sender.kyc_data.additional_kyc_data` and `Receiver.kyc_data.additional_kyc_data` respectively).
+
+The states (Sender Status, Receiver Status, Sx Flag, Rx Flag) are:
```
kphfb, line 466: Should we replace Sx and Rx references with just Sender.kyc_data.additional_kyc_data and Receiver.kyc_data.additional_kyc_data?  It feels a bit confusing to me to call them flags when they really aren't and there is no such thing as Sx and Rx they are just concepts

lips/lip-1.mdx
```diff
@@ -579,34 +688,31 @@ The `receiver_metadata_signature` will then be encoded in hex and placed into th
...
+    ...
 }
 ```
 
-For additional details, see https://lip.libra.org/lip-4/#c-to-c-transaction-flow
+For additional details, see https://dip.diem.com/dip-4/#c-to-c-transaction-flow or https://github.com/diem/client-sdk-python/blob/master/examples/p2p_transfer.py#L127
+
 
 # Reference Implementation
 A reference implementation of the Off-Chain Protocol is located at https://github.com/libra/off-chain-reference.
```
kphfb, line 718: Is this still the URL or has it been updated for diem?

LuZhang-Lou: diem => https://github.com/diem/off-chain-reference

lips/lip-1.mdx
LuZhang-Lou: Let's be explicit about whether "-" is needed here.

lips/lip-1.mdx
LuZhang-Lou: or `sender_set_additional_kyc` and `receiver_set_additional_kyc` in literal words

lips/lip-1.mdx
kphfb: @xli

lips/lip-1.mdx
```diff
@@ -492,84 +457,234 @@ Represents a national ID.
...
+
+The flows below allow the Receiver VASP and the Sender VASP to request additional KYC information from each other.
+
+### Receiver soft-match (SINIT -> RSOFT -> SSOFTSEND)
+
+After the Sender initiates a payment by providing KYC information (SINIT), the Receiver may determine they require more information to disambiguate a match. In that case they commit a command to set `Receiver.Status.Status` to `soft-match` (state RSOFT) (Step 5) or abort (step 9). The sender may respond with a command that populates the `Sender.additional_kyc_data` (Step 6), which sets the flag `Sender.additional_kyc_data` (state SSOFTSEND) or abort (SABORT, Step not shown in diagram). Finally, if the Receiver is satisfied with the additional information they move to provide all information necessary to go to RSEND (Step 4, see above. This includes receiver KYC information and recipient signature). Otherwise, the receiver can abort to state RABORT (Step 8).
```
kphfb, line 556:
```suggestion
After the Sender initiates a payment by providing KYC information (SINIT), the Receiver may determine they require more information to disambiguate a match. In that case they commit a command to set `Receiver.Status.Status` to `soft-match` (state RSOFT) (Step 5) or abort (step 9). The sender may respond with a command that populates the `Sender.additional_kyc_data` (Step 6), which sets the field `Sender.additional_kyc_data` (state SSOFTSEND) or abort (SABORT, Step not shown in diagram). Finally, if the Receiver is satisfied with the additional information they move to provide all information necessary to go to RSEND (Step 4, see above. This includes receiver KYC information and recipient signature). Otherwise, the receiver can abort to state RABORT (Step 8).
```

lips/lip-1.mdx
```diff
@@ -11,156 +11,136 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-    * An example URL: http://127.0.0.1:7091/v1/lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m/lbr1pgfpyysjzgfpyysjzgfpyysjzggqqqqqqqqqqqqqxjt54z/command/
+All HTTP requests must contain:
 
+* A header `X-REQUEST-ID` with a unique UUID (according to [RFC4122](https://tools.ietf.org/html/rfc4122)) for the request, used for tracking requests and debugging. Responses must have the same string in the `X-REQUEST-ID` header value as the requests they correspond to.
+* A header `X-REQUEST-SENDER-ADDRESS` with the HTTP request sender's VASP DIP-5 address used in the command object. The HTTP request sender must use the compliance key of the VASP account linked with this address to sign the request JWS body, and the request receiver uses this address to find the request sender's compliance key for verifying the JWS signature. For example: if the http request sender is the PaymentCommand receiver actor, then this value should be the receiver actor address from PaymentObject in PaymentCommand, and if the receiver actor address contains a ChildVASP onchain account, the request JWS body must be signed by the compliance key stored in the ParentVASP account of the ChildVASP account.
```
LuZhang-Lou, line 90: I'm trying to make this more clear: ignore me if you don't think it's better:

```suggestion
* A header `X-REQUEST-SENDER-ADDRESS` with the HTTP request sender's VASP DIP-5 address used in the command object. The HTTP request sender must use the compliance key of the VASP account linked with this address to sign the request JWS body, and the request receiver uses this address to find the request sender's compliance key to verify the JWS signature. For example: VASP A transfers funds to VASP B. In the HTTP request A sends to B, it contains `X-REQUEST-SENDER-ADDRESS` as VASP A's address. In HTTP request B sends to A, it contains `X-REQUEST-SENDER-ADDRESS` as VAPS B's address. 
```

lips/lip-1.mdx
```diff
@@ -11,156 +11,136 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+The base url value should be the url without path `/<protocol_version>/command`, e.g. `https://<hostname>:<port>`.
+
+### HTTP Headers
 
-* `protocol_version` is initially 0 for this iteration of the off-chain APIs.
-* `LocalVASPAddress` and `RemoteVASPAddress` are the bech32 VASP addresses of the respective VASPs - with the subaddress set to 0's as per [LIP 5](https://lip.libra.org/lip-5/).
-    * An example URL: http://127.0.0.1:7091/v1/lbr1pg9q5zs2pg9q5zs2pg9q5zs2pgyqqqqqqqqqqqqqqspa3m/lbr1pgfpyysjzgfpyysjzgfpyysjzggqqqqqqqqqqqqqxjt54z/command/
+All HTTP requests must contain:
 
+* A header `X-REQUEST-ID` with a unique UUID (according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included) for the request, used for tracking requests and debugging. Responses must have the same string in the `X-REQUEST-ID` header value as the requests they correspond to.
```
sunmilee, line 89: Did we decide if response also needs X-REQUEST-ID?

lips/lip-1.mdx
sunmilee: Should link "DIP-5 address" with the DIP spec

sunmilee: "In HTTP request B sends to A,..." -> "An HTTP request B sends to A should contain VASP B's address as `X-REQUEST-SENDER-ADDRESS`.

lips/lip-1.mdx
```diff
@@ -492,84 +458,235 @@ Represents a national ID.
...
+* `Sender.Status.Status`: `abort`
+* `Sender.Status.abort_code`: `rejected`.
+
+## Steps of the protocol: The soft-match states and flows
+
+A _soft-match_ occurs when a VASP checks provided KYC, but cannot disambiguate whether a party to the payment is potentially high-risk or sanctioned, just with the information provided. This may occur because the identifiers used (names, dates of birth, addresses) may partially match and are noisy. In such cases the VASP may require further information from the other VASP to process the payment, some of which may require manual intervention and interaction.
+
+The flows below allow the Receiver VASP and the Sender VASP to request additional KYC information from each other.
+
+### Receiver soft-match (SINIT -> RSOFT -> SSOFTSEND)
```
sunmilee, line 556: Do we write anywhere how long a VASP should expect to wait for a soft match response?

lips/lip-1.mdx
kphfb: Yeah, we are leaving it for now.  Xiao says it's useful for debugging purposes and we are going to ask George about it when he returns

lips/lip-1.mdx
kphfb: Updated

lips/lip-1.mdx
kphfb: Updated to include DIP 5 link.  Updated to use wording on request sender address header

kphfb: Updated to say that soft matches are suggested to be cleared within 24 hours

kphfb (approved):

lips/lip-1.mdx
```diff
@@ -11,156 +11,136 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-# Terminology
-**Object**: Equivalent to a record in a database.  A payment is an example of an Object.
+## Overview
 
-**Command**: An instruction sent over a Channel to mutate/create one or more Objects.  In the case of a mutation, this Command will depend upon the current value of one of more existing Shared Objects.
+Two VASPs participate in the off-chain protocol. They communicate through HTTP requests and responses protected with TLS, and exchange messages that are signed to ensure authenticity and integrity. The goal of the off-chain protocol is for a pair of VASPs to jointly create a `PaymentObject` containing a Reference ID. When a `PaymentObject` is **completed**, both VASPs have identical copies of the object and its Reference ID. A completed `PaymentObject` can be submitted by the Blockchain to settle.
```
dahliamalkhi, line 50:
```suggestion
Two VASPs participate in the off-chain protocol. They communicate through HTTP requests and responses protected with TLS, and exchange messages that are signed to ensure authenticity and integrity. The goal of the off-chain protocol is for a pair of VASPs to jointly create a `PaymentObject` containing a Reference ID. When a `PaymentObject` is **completed**, both VASPs have identical copies of the object and its Reference ID. A completed `PaymentObject` can be submitted to the Blockchain to settle.
```

lips/lip-1.mdx
```diff
@@ -11,156 +11,136 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-**Version**: The state of an Object at a specific point in time.  Every creation or mutation creates a new Version of the Object which was created/mutated.  VASPs operate upon the latest Version of Shared Objects.  Once a Version is mutated, it is no longer a valid Version upon which a new Command may depend - the newly updated value of the Version is the latest Version for that Object and must be acted upon.
+A VASP initially creates a `PaymentObject`, and then VASPs take turns updating it, through successful `PaymentCommands`, until it is ready to be settled or aborted. In a typical protocol flow, the VASP representing the sender would define the `PaymentObject` and include the sender's KYC information. The receiver VASP would check this information and either request more (see `soft-match` flows), abort, or signal that it is ready to settle the payment by updating sending the receiver's KYC information. The sender's VASP would check the receiver's KYC, and request more, abort, or settle the payment on-chain.
```
dahliamalkhi, line 54:
```suggestion
A VASP initially creates a `PaymentObject`, and then VASPs take turns updating it, through successful `PaymentCommand`s, until it is ready to be settled or aborted. In a typical protocol flow, the VASP representing the sender would define the `PaymentObject` and include the sender's KYC information. The receiver VASP would check this information and either request more (see `soft-match` flows), abort, or signal that it is ready to settle the payment by sending the receiver's KYC information. The sender's VASP would check the receiver's KYC, and request more, abort, or settle the payment on-chain.
```

dahliamalkhi (approved):
Great work!

Left two minor nits to handle.

One question (I may be off track here): Would it make sense for the sample code for tx submission to refer to an existing PaymentActionObject which the protocol formed and shared, and thus link back to the explanation?

lips/lip-1.mdx
```diff
@@ -11,156 +11,136 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* VASP B received both requests, and processed them in parallel. However the first request that VASP A considered as failure acquired the object / data lock by reference id.
+* VASP B processed the first request successfully and replied with a successful response, but VASP A didn’t receive it.
+* VASP B responded with a protocol error for the second request as it can’t acquire the object / data lock by reference id.
+* VASP A received failure for the second request, later VASP A may re-send the same request, and VASP B may respond with a success because the request was processed successfully.
 
 # Request/Response Payload
-All requests between VASPs are structured as [`CommandRequestObject`s](#commandrequestobject) and all responses are structured as [`CommandResponseObject`s](#commandresponseobject).  The resulting request takes a form of the following (prior to JWS):
+All requests between VASPs are structured as [`CommandRequestObject`s](#commandrequestobject) and all responses are structured as [`CommandResponseObject`s](#commandresponseobject).  The resulting request takes a form of the following (prior to JWS signing):
```
sunmilee, line 136: FYI "[`CommandRequestObject`s](#commandrequestobject)" messes up formatting: https://github.com/gdanezis/lip/blob/lip-1-update/lips/lip-1.mdx#requestresponse-payload

kphfb: thanks, fixed

PR #95, opened 2020-12-03
[Lip 10] Introduce event definitions

andll:


andll:
cc @davidiw  - can't assign you as a reviewer, can you self assign?

cc @sblackshear @tzakian  - can you take a look at events structure to make sure it makes sense?

cc @kphfb @dahliamalkhi @LuZhang-Lou

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
 * The Libra ID domain can be created as part of creating a `ParentVASP` account.
 * An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Libra IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
 
+## Libra ID domain events
+
+Blockchain contract that manages Libra ID domains must emit an event every time Libra ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Libra ID domains.
+
+While Libra ID domains published into VASP accounts, Libra ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
```
LuZhang-Lou, line 82: Is this the same approach we are applying to compliance key/ base url update events? Cc @tzakian

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
...
+```
+resource struct LibraIdDomainManager {
+    /// Events are emmitted 
+    libra_id_domain_events: Event::EventHandle<Self::LibraIdDomainEvent>,
+}
+
+struct LibraIdDomainEvent {
+    removed: bool,
+    domain: LibraIdDomain,
+    address: address,
```
LuZhang-Lou, line 95: If it is a domain id update event, we need to know its old domain id, because a parent Vasp can have multiple of domain ids

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
...
+}
+```  
+
+* Fields definition:
+  * `removed` - `true`, if LibraIdDomain was removed, `false` if it was created or updated
+  * `domain` - exact copy of `LibraIdDomain` that was added/removed from `LibraIdDomains` resource of a VASP account
+  * `address` - address of an account where `LibraIdDomain` was added or removed
```
LuZhang-Lou, line 102: Just to clarify here the address here is a parent Vasp address?

lips/lip-10.md
andll: There is no really 'id' of a domain.
The name of domain can be used to identify it, it is part of `LibraIdDomain`.

lips/lip-10.md
andll: Yes

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
 * The Libra ID domain can be created as part of creating a `ParentVASP` account.
 * An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Libra IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
 
+## Libra ID domain events
+
+Blockchain contract that manages Libra ID domains must emit an event every time Libra ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Libra ID domains.
```
davidiw, line 80: Rewrite this and the following paragraph:

The blockchain contract that manages Libra ID domains emits an event every time a Libra ID domain is created, removed, or updated. These events are published under the Treasury Compliance account, whereas the actual Libra ID domains are published under VASP accounts. Applications can efficiently index Libra ID domains by querying these events conveniently from the Treasury Compliance account.

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
...
+
+While Libra ID domains published into VASP accounts, Libra ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
+
+To store events, `LibraIdDomainManager` resource is published under the Treasure Compliance account(address `0xB1E55ED`).
```
davidiw, line 84: The `LibraIdDomainManager` resources acts as an event store and is published under the ...

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
...
+
+```
+resource struct LibraIdDomainManager {
+    /// Events are emmitted 
+    libra_id_domain_events: Event::EventHandle<Self::LibraIdDomainEvent>,
+}
+
+struct LibraIdDomainEvent {
+    removed: bool,
```
davidiw, line 93: recommend we have a "enum" (u8)

0 - created
1 - modified
2 - removed

lips/lip-10.md
```diff
@@ -73,6 +75,32 @@ struct LibraIdDomain {
...
+    domain: LibraIdDomain,
```
davidiw, line 94: what do we benefit by having the actual value here? this seems to be a potential waste of resources.

davidiw (commented):
overall looks good, thanks for adding this section, minor nits and discussion around what we actually want in the event itself

lips/lip-10.md
LuZhang-Lou: I think my question is:
let's say we update a domain for an address, how do we know the old value that is updated? 
For example, when Novi updates one of its domains "www.novi.com" to "www.novi2.com", emitting the following event does not let people know the old domain (e.g. whether it is updated from "www.novi.com" or "www.novi_sth_else.com" ) 
```
struct LibraIdDomainEvent {
    removed: false,
    domain: "www.novi2.com,
    address: "some_address",
 }
```

lips/lip-10.md
LuZhang-Lou: I am wondering about the same thing. It's likely that we need to add another field for old domain on update cases. In that case, maybe we do not reveal the values but just an event that tells people this address has an update for libraid domain. The trade-off is people will need to look it up by themselves.

andll: So there are basically two alternatives:
- Pass entire value like we do right now
- Pass some 'incremental update' struct that has array of (field, value)

Second option seem to be less than ideal as the contract will have to figure out how to generate it, and it is also not going to be strictly typed (e.g. what type `value` part of such incremental update should be?)

I think if we had large frequently updated structure I could understand argument of saving resources by not providing full value, but in current circumstances it seem pretty natural choice to just put 'latest value' here

lips/lip-10.md
andll: Oh, I see

I don't think we going to support simple operation of 'updating address'.

In practice, updating address is going to be multi-stage process, that basically consists of introducing new domain, and slowly deprecating old domain, and then finally removing it from the chain

lips/lip-10.md
LuZhang-Lou: Another way is not to have any values (even incremental ones) in the event. The only message the event conveys is "this address has an update with libra id domains".

lips/lip-10.md
kphfb: In the domain, Lu shows it as www.novi2.com, but wouldn't it be more like just "novi"?

And if there is no updating, then why do we even have update as a type in the "op"?

lips/lip-10.md
kphfb: I had the same question as Lu.   Can we add that this is the parent address and example values also?

lips/lip-10.md
```diff
@@ -73,6 +75,33 @@ struct LibraIdDomain {
...
+  * `op` - indicates whether Libra ID domain was:
+     * `0` - created
+     * `1` - modified
+     * `2` - removed
+  * `domain` - exact copy of `LibraIdDomain` that was added/removed from `LibraIdDomains` resource of a VASP account
+  * `address` - address of an account where `LibraIdDomain` was added or removed
+
 # Off-chain protocol
 
 **TODO: This section may change in the future as we update LIP 1**
```
kphfb, line 107: We've updated LIP 1 now, so you can update this LIP as well

lips/lip-10.md
sblackshear: What does `modified` mean?

lips/lip-10.md
sblackshear: What @andll has here is consistent with the style we use in other events. Elsewhere, we try to make events self-contained so you can understand them without needing to look at the associated state/transactions.

In this case, I think it makes sense to include the domain in the event. If you do that, the domain indexer will only need to look at the `LibraIdDomainEvent` stream to know the complete set of available domains + their owners.

lips/lip-10.md
andll: >In the domain, Lu shows it as www.novi2.com, but wouldn't it be more like just "novi"?

Yes, currently domain is something like 'novi', domain regex does not allow `.` in it.

>And if there is no updating, then why do we even have update as a type in the "op"?

There is no updating of domain name(e.g. identifier), but we most likely going to have more fields in `LibraIdDomain`, which can be updated

lips/lip-10.md
andll: Yeah, I still think having value is much easier for indexer to process.

lips/lip-10.md
```diff
@@ -50,6 +50,8 @@ Example: `andrey$novi`
   * Maximum length: 63 characters
   * Association between libra id and VASP is published on the chain 
 
+#On-chain API
+
 ## Libra ID domains
```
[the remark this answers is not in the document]
andll: Yes, but perhaps separate PR is better for this

lips/lip-10.md
```diff
@@ -50,6 +50,8 @@ Example: `andrey$novi`
   * Maximum length: 63 characters
   * Association between libra id and VASP is published on the chain 
 
+#On-chain API
```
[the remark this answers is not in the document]
andll: Renamed to 'On-chain data'

dips/dip-10.md
```diff
@@ -73,6 +75,32 @@ struct DiemIdDomain {
 * The Diem ID domain can be created as part of creating a `ParentVASP` account.
 * An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
 
+## Diem ID domain events
+
+Blockchain contract that manages Diem ID domains must emit an event every time Diem ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Diem ID domains.
+
+While Diem ID domains published into VASP accounts, Diem ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
```
davidiw, line 82: Diem ID domains *are* published into VASP accounts

davidiw: Do we explicitly state what this address is anywhere? I'm curious how folks would find it.

dips/dip-10.md
```diff
@@ -73,6 +75,32 @@ struct DiemIdDomain {
 * The Diem ID domain can be created as part of creating a `ParentVASP` account.
 * An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
 
+## Diem ID domain events
+
+Blockchain contract that manages Diem ID domains must emit an event every time Diem ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Diem ID domains.
```
davidiw, line 80: *The Move* contract

davidiw (approved):
just a couple language nits and then let's land it -- I think I might have to land it as well, let me know

dips/dip-10.md
andll: Yes, this document mentions specific address (I can repeat it here too, but it feels redundand)

PR #99, opened 2020-12-04
Rename Libra -> Diem

valerini:
Rename:
* Libra -> Diem,
* LN -> DN,
* LA -> DA,
* LPN -> DPN,
* LIP -> DIP.

Problems:
* The links still point to libra to keep them valid.
* The names of the following resources are incorrect now (do not match the codebase): DiemSystem, DiemConfig.

### DIFF
--- a/lips/lip-6.mdx
+++ b/lips/lip-6.mdx
@@ -1,6 +1,6 @@
 ---
-lip: 6
-title: Libra Validator Configuration Management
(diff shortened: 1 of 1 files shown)


davidiw (approved):
we'll slowly be witnessing a name change in the code base, we should be comfortable living in this quirky world for now :)

PR #105, opened 2020-12-18
Update dip-8.md to match LIP 1 updates

kphfb:
Updating LIP 8  to match the changes to LIP-1 as per version 2 of LIP-1 which  simplified the flow of LIP-1

### DIFF
--- a/dips/dip-8.md
+++ b/dips/dip-8.md
@@ -1,6 +1,6 @@
 ---
 dip: 8
-title: Off-chain API Version 1 Extensions
+title: Off-chain API Version 2 Extensions
 authors: Kevin Hurley (@kphfb)
 status: Draft
 type: Informational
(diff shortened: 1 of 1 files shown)


dips/dip-8.md
```diff
@@ -15,14 +15,14 @@ An extension of the Off-Chain protocol to provide support for more advanced merc
...
-Version 0 of the Off-Chain Protocol is described in [DIP 1](https://dip.diem.com/dip-1/).  Version 1 as described here is an extension of the Off-Chain Protocol and adds features to support more advanced functionality - particularly targeted to support merchant use-cases.  This is inclusive of pull payments, recurring payments, and auth/capture flows.
+Versions 1/2 of the Off-Chain Protocol are described in [DIP 1](https://dip.diem.com/dip-1/).  Version 3 as described here is an extension of the Off-Chain Protocol and adds features to support more advanced functionality - particularly targeted to support merchant use-cases.  This is inclusive of pull payments, recurring payments, and auth/capture flows.
 
 ---
 # Specification
 ---
 
 # Fund Pull Pre-Approval
-### *VERSION SUPPORT: Supported only in version 1 of off-chain APIs*
+### *VERSION SUPPORT: Supported only in version 3 of off-chain APIs*
```
xli, line 25: is this version 3 match the version number in API path? version 2 uses v2, this is v3?

kphfb: Yeah, that's what I was thinking

LuZhang-Lou: I was confused by this because at the moment we don't have v3 of that. Does it make more sense to say "This is not yet supported in off chain APIs but will soon be" ?

xli: There is no way to understand what version a VASP is supporting, so it's all or none, if some VASPs implemented v3, how they may communicate with others? some time use v2, some time use v3 ?

xli: Let me give an example, for vasp a, and vasp b:
1. vasp a calls vasp b for fund pre-approval, url: <base_url>/v3/command
2. vasp a calls vasp b for payment object referenced to the fund pre-approval, url: <base_url>/v3/command
3. now if there is payment from vasp a to vasp b without fund pre-approval, what url it should call? <base_url>/v3? or v2?

davidiw:
I think this type of work is really exciting and would love to dig into it!

With that being said,  I want to begin torturing Kevin:

* Need more contents in abstract than “This is inclusive of pull payments, recurring payments, and auth/capture flows.”
* goal / problem statement for fund pre-approval isn’t very clear, needs a better strawman example or just being broken down specifically what each protocol is trying to attempt (I like bullet points) -- for example, what is the properly end-to-end flow involving outside -- should I as a consumer pre-authorize something, will this appear in my account to confirm, it feels like something is missing here.
* Why does Auth/Capture leverage PaymentCommand versus another command? We would likely still need to perform a KYC prior to TR transactions anyway and we can leverage the slightly new style of performing these types of request / responses (no need for an explicit callback can handle in a single http request / response)
* If we use new commands, then the “version” of the API is irrelevant and instead it is whether or not certain commands are supported -- in fact -- we should really figure out a versioning for objects (maybe just different names) than versioning the API because the API should only change if we adjust the CommandObject format or change the format of the lower level protoocols.
* The usage of headers is inconsistent # defines specifications and several fields within specifications (e.g., Auth/Capture), children then use ###, we’re skipping ##, the madness

kphfb:
Oh, well good news is that @danprinz from First is going to re-work this DIP.  And split it out into two as well

davidiw:
> Oh, well good news is that @danprinz from First is going to re-work this DIP. And split it out into two as well

Good, I will bother him appropriately on his PR too then <3

davidiw:
feel free to re-open but this is pretty stale now

PR #114, opened 2021-01-05
nits that came as a result of recreating the dips repo from lip-placeholder

oriooctopus:


JoelMarcey (approved):
We probably want to also delete the `gh-pages` branch at some point after we do the redirect changes necessary for lip.libra.org. cc @jjwill

PR #117, opened 2021-01-08
add error code list to dip-1

xli:


dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+* Field `recipient_signature` value is not hex-encoded bytes.
+* Field `recipient_signature` value is an invalid signature.
+
+`unknown_address`:
+* The DIP-5 account identifier address in the command object is not HTTP request sender’s address or receiver’s address. For payment object it is `sender#address` or `receiver#address`.
+* Could not find on-chain account by an DIP-5 account identifier address in command object address.
+
+`conflict`:
+* Command object is in conflict with another different command object by cid, likely a cid is reused for different command object.
+* Failed to acquire lock by for the command object by the `reference_id`.
```
sunmilee, line 275:
```suggestion
* Failed to acquire lock for the command object by the `reference_id`.
```

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+`unsupported_currency`: field `payment#action#currency` value is a valid Diem currency code, but it is not supported / acceptable by the receiver VASP.
```
sunmilee, line 277: Do you get a different error if it's not a valid currency?

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+##### Request Object Validation Error Codes
+
+`invalid_json`: decoded JWS body is not json.
+
+`invalid_object`: command request/response object json is not object, or the command object type does not match `command_type`.
+
+`missing_field`:
+* Missing required field.
+* An optional field is required to be set for a specific state, e.g. PaymentObject requires sender's `kyc_data` (which is an optional field for `PaymentActorObject`) when sender init the `PaymentObject`.
```
gdanezis, line 247: Do we communicate in any way which field is missing?

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+`unknown_field`: field is unknown for an object.
```
gdanezis, line 249: Do we communicate back the offending field? How strict are we in general accepting fields we do not understand? One approach to extensibility is to tolerate unknown fields, so that extesions to the LIP-1 do not interfere with older implementations.

dips/dip-1.mdx
gdanezis, line 275: This failed to acquire lock confuses me: now we do not have sender / receiver , client / server competing for objects? So when can a lock fail?

LuZhang-Lou: we should remove this

dips/dip-1.mdx
sunmilee: Think there is a string error msg you can write any info into -- can we add this here?

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+* Invalid / unknown enum field values.
+* UUID field value does not match UUID format.
+* Payment actor address is not a valid DIP-5 account identifier.
+* Currency field value is not a valid Diem currency code for the connected network.
+
+`invalid_command_producer`: The HTTP request sender is not the right actor to send the payment object. For example, if the actor receiver sends a new command with payment object change that should be done by actor sender.
+
+`invalid_initial_or_prior_not_found`: could not find command by reference_id for a non-initial state command object; for example, actor receiver received a payment command object that actor sender status is `ready_for_settlement`, but receiver could not find any command object by the reference id.
+
+`no_kyc_needed`: payment action amount is under travel rule limit.
```
sunmilee, line 263: Will this actually be used? Whoever triggers a travel rule interaction is probably doing it because they want KYC info, whether it's above or below threshold?

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
     "type": "command_error",
     "field": "payment.sender.kyc_data.surname",
-    "code": "missing_data",
+    "code": "missing_field",
     "message": "",
 }
 ```
-`command_error` occurs in response to a Command failing to be applied - for example, a high level validation error.  `protocol_error` occurs in response to a failure related to the lower-level protocol.  Note: Additional information detailing all errors will be added.
 
+`command_error` occurs in response to a Command failing to be applied - for example, a high level validation error.  `protocol_error` occurs in response to a failure related to the lower-level protocol.
```
LuZhang-Lou, line 219: let's be more explicit with these errors: 
command_error results committed failed commands, but commands with protocol_error are retry-able.

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+#### List of Error Codes
+
+##### HTTP Header Validation Error Codes
+
+`invalid_x_request_sender_address`:
+* `X-REQUEST-SENDER-ADDRESS` header value is not the request sender’s address in the command object. All command objects should have a field that is the request sender’s address. For payment object, it is `sender#address` or `receiver#address`.
```
LuZhang-Lou, line 226: sender#address -> sender.address, same to receiver

that's the format we use throughout this doc

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+* Could not find Diem's onchain account by the `X-REQUEST-SENDER-ADDRESS` header value.
+* Could not find the compliance key of the onchain account found by the `X-REQUEST-SENDER-ADDRESS` header value.
+* The compliance key found from the onchain account is not a valid ED25519 public key.
```
LuZhang-Lou, line 229: I feel these two sound more like vasp setup error, v.s. header error, maybe worth a separate error code?

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+`missing_http_header`: missing HTTP header `X-REQUEST-ID` or `X-REQUEST-SENDER-ADDRESS`.
+
```
LuZhang-Lou, line 232: and malformed x-request-id (not uuid)?

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+`invalid_initial_or_prior_not_found`: could not find command by reference_id for a non-initial state command object; for example, actor receiver received a payment command object that actor sender status is `ready_for_settlement`, but receiver could not find any command object by the reference id.
+
+`no_kyc_needed`: payment action amount is under travel rule limit.
+
+`invalid_recipient_signature`:
+* Field `recipient_signature` value is not hex-encoded bytes.
+* Field `recipient_signature` value is an invalid signature.
+
+`unknown_address`:
+* The DIP-5 account identifier address in the command object is not HTTP request sender’s address or receiver’s address. For payment object it is `sender#address` or `receiver#address`.
```
LuZhang-Lou, line 270: same for`#` -> `->`

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+* Could not find on-chain account by an DIP-5 account identifier address in command object address.
+
+`conflict`:
+* Command object is in conflict with another different command object by cid, likely a cid is reused for different command object.
+* Failed to acquire lock by for the command object by the `reference_id`.
+
+`unsupported_currency`: field `payment#action#currency` value is a valid Diem currency code, but it is not supported / acceptable by the receiver VASP.
+
+`missing_dependency`: One of dependency objects is not found by the id provided in the command object. For example, the receiver should return this error code if it could not find the data by the `original_payment_reference_id` if the sender set it.
```
LuZhang-Lou, line 279: `dependency` is the concept when we still have versioning, since this is gone, we'd better remove all related errors. Regarding `original_payment_reference_id`, a separate error code is worth it :)

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+`invalid_dependency`: One of dependency objects is found, but it is not valid for the command object. For example, the dependency payment object status is `aborted` instead of `ready_for_settlement`.
```
LuZhang-Lou, line 281: this is covered by "invalid_transition"

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
+
+`invalid_overwrite`:
+* Overwrite a field that can only be written once.
+* Overwrite an immutable field (field can only be set in initial command object), e.g. `original_payment_reference_id`).
+* Overwrite opponent payment actor's fields.
+
+`invalid_transition`: As we only allow one actor action at a time, and the next states for a given command object state are limited to specific states. This error indicates the new payment object state is not valid according to the current object state. For example: VASP A sends RSOFT to VASP B, VASP B should send the next payment object with ABORT, or SSOFTSEND; VASP A should respond to this error code if VASP B sends payment object state SSOFT.
```
LuZhang-Lou, line 288: I think this actually covers `invalid_command_producer`

dips/dip-1.mdx
```diff
@@ -211,18 +211,88 @@ Represents an error that occurred in response to a Command.
...
-    "code": "missing_data",
+    "code": "missing_field",
     "message": "",
 }
 ```
-`command_error` occurs in response to a Command failing to be applied - for example, a high level validation error.  `protocol_error` occurs in response to a failure related to the lower-level protocol.  Note: Additional information detailing all errors will be added.
 
+`command_error` occurs in response to a Command failing to be applied - for example, a high level validation error.  `protocol_error` occurs in response to a failure related to the lower-level protocol.
+
+#### List of Error Codes
```
LuZhang-Lou, line 221: how do we support business errors? Cover all with a "rejected" ?

dips/dip-1.mdx
LuZhang-Lou, line 247: can we have a separate error for state specific errors, e.g. 
SINIT has no sender.kyc, and RSEND has no receiver.kyc or recipient_signature. 
The ration behind that is since we introduce the concepts of states in new dip1 (SINIT, RSEND etc), it's natural to check the validity of states, hence an error code is worth it

dips/dip-1.mdx
LuZhang-Lou: can we specify their types of every error below?

dips/dip-1.mdx
xli: invalid-field-value

dips/dip-1.mdx
xli: see OffChainErrorObject#field

dips/dip-1.mdx
xli: For example: 
* VASP A sends sinit to VASP B
* VASP B side has a bug, it sends 2 commands at almost same time:
  *  rsend
  *  rsoft
*  VASP B received the command requests, if it does not lock the previous sinit before start to validate and save the command, then 2 requests are processed, one of them overwrites another one, the result state is unknown.
* VASP B received the success responses, and save the result, one of them will be saved.
* It is possible we ende with one of the following result: 
  *  VASP A: rsend, VASP B: rsoft
  * VASP A: rsoft, VASP B: rsend
* Once we enter the above inconsistent result, VASP A's following up command request will always fail.

dips/dip-1.mdx
xli: The error codes are for the validation result of request receiver, it can't assume sender has no bug.
From receiver's point view, it should avoid unnecessary kyc data exchange which may cause data leak.

dips/dip-1.mdx
xli: Added to top about error code and validations are optional depending on implementation.

dips/dip-1.mdx
xli: The error code is http request receiver's point view, it does not need care why it is wrong,

dips/dip-1.mdx
xli: changed invalid_x_request_sender_address to invalid_http_header to cover both cases.

dips/dip-1.mdx
xli: I think dependency makes sense for original_payment_reference_id and future cases.

dips/dip-1.mdx
xli: updated the example description, it is not covered by invalid_transition.

dips/dip-1.mdx
xli: it is not, a command object with invalid_command_producer can be valid transition. An invalid_command_producer is more of validating the command object with the x_request_sender_address.

dips/dip-1.mdx
xli: If you are talking about set actor.status.status to abort with code reject, then no, because "rejected" is the code the status abort.

dips/dip-1.mdx
xli: missing_state_field? I could not find a good name for it

xli: Also OffChainErrorObject has field named "field" for what is field missing, hence missing_kyc_data is not necessary.

dips/dip-1.mdx
xli: I'm not up to mark anything "retry-able". 
All errors are all retry-able, for example, for `missing_field`: it is possible server side introduced a bug to raise the error, once error rates increased, the deployment of bug code is reverted, then client side retry can resolve the error and move things forward.

dips/dip-1.mdx
LuZhang-Lou: > a command object with invalid_command_producer can be valid transition

can you give an example? Note that only the sender can change sender's status, same to receiver.

xli: Let's say sender sends SINT, then sender sends another new command RSEND, with all information set for receiver. It changes receiver actor object, which is invalid-overwrite-error, but it is also invalid_command_producer, and the result is a valid transition.

It probably related to what is a scope of "transition", from my point view, it is from command object A to B, regardless of who created B from A. But if you consider what actor create B from A is also part of transition validation, then yes, it is covered.
It's common some error codes may cover each other in a complex scenario, we don't need judge what and when must be used.
For example, technically json `"{}" ` can trigger a lot of errors, like invalid transition, invalid overwrite, missing fields..... Application just need try to return reasonable one for it.

dips/dip-1.mdx
xli: I mean, the request receiver has no idea whether the onchain account setup is the root of the problem, or it is because sender sends wrong x-request-sender-address. Hence it should not assume the specific error happened.

dips/dip-1.mdx
xli: Or missing_data?

dips/dip-1.mdx
LuZhang-Lou: > Let's say sender sends SINT, then sender sends another new command RSEND, with all information set for receiver. It changes receiver actor object, which is invalid-overwrite-error, but it is also invalid_command_producer, and the result is a valid transition.

SENDER shouldn't be allowed to send a `RSEND` command and it can be interpreted from two different aspects:
1. to get to RSEND, sender needs to touch receiver's data, this is not allowed (they shall only update their own data, respectively)
2. SENDER should never reach RSEND, because it is a state that only receiver can move into.

But this is fair enough, no strong opinion over one or the other

dips/dip-1.mdx
LuZhang-Lou: `unrecognized_original_reference_id`?

dips/dip-1.mdx
xli: You're right about sender should never reach RSEND, I'm giving an example of invalid/buggy implementation that may give "a command object with invalid_command_producer can be valid transition"

dips/dip-1.mdx
xli: updated as invalid_original_reference_id and merged invalid_dependencies too.

LuZhang-Lou (approved):

dips/dip-1.mdx
```diff
@@ -556,11 +629,11 @@ The flows below allow the Receiver VASP and the Sender VASP to request additiona
...
 
 ### Sender soft-match (RSEND -> SSOFT -> RSOFTSEND)
 
-Similarly to the flow above, the Sender at state (RSEND) may decide they need information about the receiver of the payment to disambiguate a soft-match. They set their status `Sender.Status.Status` to `soft-match` (state SSOFT) (Step 5). The receiver can abort (RABORT) (Step not shown in diagram) or provide the additional KYC in `Receiever.additional_kyc_data` (state RSOFTSEND, Step 6). The sender may then abort (SABORT, Step 8) or move to READY (Step 7), and settle the payment.
+Similarly to the flow above, the Sender at state (RSEND) may decide they need information about the receiver of the payment to disambiguate a soft-match. They set their status `sender.status.status` to `soft-match` (state SSOFT) (Step 5). The receiver can abort (RABORT) (Step not shown in diagram) or provide the additional KYC in `receiever.additional_kyc_data` (state RSOFTSEND, Step 6). The sender may then abort (SABORT, Step 8) or move to READY (Step 7), and settle the payment.
```
kphfb, line 636:
```suggestion
Similarly to the flow above, the Sender at state (RSEND) may decide they need information about the receiver of the payment to disambiguate a soft-match. They set their status `sender.status.status` to `soft-match` (state SSOFT) (Step 5). The receiver can abort (RABORT) (Step not shown in diagram) or provide the additional KYC in `receiver.additional_kyc_data` (state RSOFTSEND, Step 6). The sender may then abort (SABORT, Step 8) or move to READY (Step 7), and settle the payment.
```

kphfb (approved):

PR #118, opened 2021-01-08
Revisions to DIP-20

tzakian:
Updates DIP-20 based on Dahlia's comments and suggestions.


dahliamalkhi (approved):
Thanks for the massive work!

PR #120, opened 2021-01-13
Update dip-1.mdx

davidiw:
* Make the reference and cid must and instead of should, it seems a little strange that in other cases we are very clear about format and shape and in this case we offer flexibility
* clean up language like (bytes, ASCII) that's just bytes and bytes, hex is hex-encoded bytes
* subaddress is now address / participant where appropriate
* do not use uleb-128 for onchain metadata representation of the ref id since it doesn't compress it actually expands the data and just adds more complexity to this protocol

### DIFF
--- a/dips/dip-1.mdx
+++ b/dips/dip-1.mdx
@@ -18,8 +18,15 @@ The Off-Chain Protocol allows two entities to define payments off-chain and priv
 
 | Version | Revision | Link      |
 |---------|----------|-----------|
-| v2      | current  | n/a       |
-| v1      | e82a0eb8a9dd5d498fc89bf84565cc7adae3d8ef | https://github.com/diem/dip/blob/e82a0eb8a9dd5d498fc89bf84565cc7adae3d8ef/dips/dip-1.mdx  |
+| v3      | current  | n/a       |
+| v2	  | 90af9435e933a5ff46031d254acb3a52bf81f67f | https://github.com/diem/dip/blob/90af9435e933a5ff46031d254acb3a52bf81f67f/dips/dip-1.mdx  |
+| v1      | e82a0eb8a9dd5d498fc89bf84565cc7adae3d8ef | https://github.com/diem/dip/blob/e82a0eb8a9dd5d498fc89bf84565cc7adae3d8ef/lips/lip-1.mdx  |
+
+V3 changes:
+
+* Remove subaddresses and use only VASP account-addresses
+* Sender retrieves the reference id from the recipient using a DIP-4 Payment URI
+* Changed reference_id to use a 64 bit identifier so that it replaces subaddresses
 
 V2 changes:
 
@@ -69,7 +76,7 @@ The remainder of this document details the specifications for Commands, Objects 
 
 **Shared Object**: All Objects contained within a Command are logically shared, meaning that each VASP has a copy of the Object and may create a new Command to modify it. For example, during the lifecycle of a `KycDataObject`, both VASPs will add information to it. The protocol does not support simultaneous updates to Objects. Rather, VASPs must take turns updating the Object - with turns specified as per the [command sequence](#paymentobject-protocol-command-sequence).
 
-**Reference ID**: Every Object type contains a `reference_id` field which is a globally unique reference ID for the Object. The reference ID is always specified by the payment initiator VASP (the VASP which originally created the Object). The value must be a 128 bit long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included.
+**Reference ID**: Every Object type contains a 64-bit `reference_id` field. The (sender account, recipient account, and `reference_id`) is a globally unique identifier for this Object. This value should be obtained via a [Diem Account Identifier](https://dip.diem.com/dip-5/) that the sender requests from the recipient before submitting payment.
 
 ### On-Chain Account Setup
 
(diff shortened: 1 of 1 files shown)


dips/dip-1.mdx
```diff
@@ -332,7 +332,7 @@ A `PaymentActorObject` represents a participant in a payment - either sender or
 
 | Field 	    | Type 	| Required? 	| Description 	|
 |-------	    |------	|-----------	|-------------	|
-| address | str | Y | Address of the sender/receiver account. Addresses may be single use or valid for a limited time, and therefore VASPs should not rely on them remaining stable across time or different VASP addresses. The addresses are encoded using bech32. The bech32 address encodes both the address of the VASP as well as the specific user's subaddress. They should be no longer than 80 characters. Mandatory and immutable. For Diem addresses, refer to the "account identifier" section in DIP-5 for format. |
+| address | str | Y | The 16-byte on-chain address of the sender/receiver account in case-insensitive, hex encoding. |
```
LuZhang-Lou, line 335: I'm not sure about hex encoding or bech32 encoding here. Everywhere else in this dip uses bech32. Better to make them consistent.

dips/dip-1.mdx
```diff
@@ -614,24 +614,21 @@ The data that contributes to the compliance recipient signature.
 
     amount (u64) = "5123456" (Hex "802d4e0000000000")
 
-Metadata is serialized using  LCS (including encoding of `reference_id`) and appended to the fixed length byte sequences representing `address`, `amount`, and `DOMAIN_SEPARATOR`. For example the byte sequence that is signed for the transaction data above is (bytes, hex):
+Metadata is serialized using  LCS (including encoding of `reference_id`) and appended to the fixed length byte sequences representing `address`, `amount`, and `DOMAIN_SEPARATOR`. For example the byte sequence that is signed for the transaction data above is (in hex-encoded bytes):
```
LuZhang-Lou, line 617: using DCS , and an extra whitespace

dips/dip-1.mdx
```diff
@@ -614,24 +614,21 @@ The data that contributes to the compliance recipient signature.
...
 
-     "02000120626239393164386533653630313165623865616561636465343830303131323253414d504c4552454641444452455353802d4e0000000000404024244449454d5f41545445535424244040" (len=158)
+     "02004257855C8410735AC58E2B5C422C479453414d504c4552454641444452455353802d4e0000000000404024244449454d5f41545445535424244040" (len=122)
```
LuZhang-Lou, line 619: why this is changed when the key and payloads stay the same?

dips/dip-1.mdx
```diff
@@ -614,24 +614,21 @@ The data that contributes to the compliance recipient signature.
...
 
 The above serialized byte array represents:
 
         "0200"                                      - Metadata type and version (2 bytes, constant value)
-        "0120"                                      - uleb128 encoded reference_id length (variable)
```
LuZhang-Lou, line 624: I still don't get why length is removed

dips/dip-1.mdx
```diff
@@ -614,24 +614,21 @@ The data that contributes to the compliance recipient signature.
...
-        "6262 ... 3232"                             - Bytes of reference_id (variable)
+        "4257855C8410735AC58E2B5C422C4794"          - Bytes of reference_id (16 bytes)
```
LuZhang-Lou, line 624: => reference_id (16 bytes, hex-encoded)

dips/dip-1.mdx
davidiw: lemme know if you see any other bech32 in here, because we don't have a bech32 encoder / decoder for just an account address

dips/dip-1.mdx
davidiw: fixed, good call

dips/dip-1.mdx
davidiw: I kept the format that matches the rest of this, lemme know if you think we should add the hex encoding to everything thing here

dips/dip-1.mdx
davidiw: we already know the length is going to be 128-bits, why are we writing the length to the chain?

dips/dip-1.mdx
davidiw: this value needs to be tweaked to be 8bytes but the rest of the doc represents discussions thus far

dips/dip-1.mdx
```diff
@@ -69,7 +76,7 @@ The remainder of this document details the specifications for Commands, Objects
...
 
-**Reference ID**: Every Object type contains a `reference_id` field which is a unique reference ID for the Object. The reference ID is always specified by the payment initiator VASP (the VASP which originally created the Object). This value should be globally unique. It is recommended to use a 128 bit long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included.
+**Reference ID**: Every Object type contains a 64-bit `reference_id` field. The (sender account, recipient account, and `reference_id`) is a globally unique identifier for this Object. This value should be obtained via a [Diem Account Identifier](https://dip.diem.com/dip-5/) that the sender requests from the recipient before submitting payment.
```
xli, line 79: could we recommend a way to generate 64-bit reference_id?

dips/dip-1.mdx
xli: The metadata bytes are serialized by BCS, this uleb128 for length of an array, is BCS spec.
This part example describes what bytes means after BCS serialized the data. You can't change it.

dips/dip-1.mdx
```diff
@@ -86,7 +93,7 @@ The base url value should be the url without path `/<protocol_version>/command`,
 
 All HTTP requests must contain:
 
-* A header `X-REQUEST-ID` with a unique UUID (according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included) for the request, used for tracking requests and debugging. Responses must have the same string in the `X-REQUEST-ID` header value as the requests they correspond to.
+* A header `X-REQUEST-ID` with a unique 64-bit identifier for the request, used for tracking requests and debugging. Responses must have the same string in the `X-REQUEST-ID` header value as the requests they correspond to.
```
xli, line 96: this needs to be UUID. it's for debug purpose, and need to be global uniq

dips/dip-1.mdx
```diff
@@ -163,7 +170,7 @@ All requests between VASPs are structured as `CommandRequestObject`s.
 | _ObjectType  | str    | Y | Fixed value: `CommandRequestObject`|
 | command_type | str    | Y | A string representing the type of Command contained in the request. |
 | command      | Command object | Y | The Command to sequence. |
-| cid | str    | Y     | A unique identifier for the Command. Should be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included. |
+| cid | str    | Y     | A unique 64-bit identifier for the Command. |
```
xli, line 173: Why cid is changed? This is uniq id for the command object.

dips/dip-1.mdx
xli: One question about this change is the HRP of the account identifier was used for checking the network of the payment. The account address may belongs to multiple network, this maybe confusing to sender/receiver what network sender is going to submit the transaction.

aching:
Is this PR still active, or is the thought to create a new DIP that obsoletes DIP-1?

davidiw:
> Is this PR still active, or is the thought to create a new DIP that obsoletes DIP-1?

closing out will update with something shortly

PR #121, opened 2021-01-13
[DIP-1] disallow partial updates

LuZhang-Lou:
Partial update (i.e. not sending a complete command but only a diff) is not super useful but increases the amount of implementation work for all VASPs/DDs. Luckily we did not mention the concept of partial updates a lot in the DIP. So we just remove the parts indicating that is allowed.


dips/dip-1.mdx
```diff
@@ -303,13 +303,13 @@ For a travel rule data exchange, the [command_type](#commandrequestobject) field
...
 
 | Field 	    | Type 	| Required? 	| Description 	|
 |-------	    |------	|-----------	|-------------	|
-| sender/receiver | [`PaymentActorObject`](#paymentactorobject) | Required for payment creation | Information about the sender/receiver in this payment |
+| sender/receiver | [`PaymentActorObject`](#paymentactorobject) | Y | Information about the sender/receiver in this payment |
 | reference_id | str | Y | Unique reference ID of this payment on the payment initiator VASP (the VASP which originally created this payment Object). This value should be globally unique. This field is mandatory on payment creation and immutable after that. We recommend using a 128 bits long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included. |
-| original_payment_reference_id | str | N | Used for updates to a payment after it has been committed on chain. For example, used for refunds. The reference ID of the original payment will be placed into this field. This value is optional on payment creation and invalid on updates. |
```
xli, line 312: so the value can be updated after creation?

LuZhang-Lou: good call out, I should note it's immutable once populated. But first let me confirm the correct use of this field in refund flow with KH

xli (approved):

kphfb (approved):

PR #126, opened 2021-01-14
Fix Algolia Search

JoelMarcey:
We were using the wrong API key for search.  So no results. This fixes the API key.


oriooctopus (approved):
LGTM

PR #130, opened 2021-01-21
[dip5] update hrps and explicitly disallow mixed-cases

LuZhang-Lou:
1. update hrps 
2. explicitly disallow mixed-cases for diem intent identifier. This has been causing confusions.


kchalkias (approved):
Thanks!

kphfb (approved):

PR #133, opened 2021-01-29
Address comments 

tzakian:
Addresses a number of additional comments received on DIP-20

### DIFF
--- a/dips/dip-20.md
+++ b/dips/dip-20.md
@@ -15,7 +15,7 @@ This DIP describes the conceptual model and implementation of currencies on
 the Diem blockchain. Currencies in Diem are represented statically at the
 type level, each currency with a unique Move type.
 
-The Diem currency implementation maintains coin safety rules through two
+The Diem Currency implementation maintains coin safety rules through two
 principal mechanisms of the Move language, permissions and types.
 Permissions guarantee coin scarcity, in that a coin can be created
 ("minted") or destroyed ("burned") only through explicit "treasury"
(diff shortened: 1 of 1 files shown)


dips/dip-20.md
```diff
@@ -124,44 +123,42 @@ account address (see the implementation of
...
+at the time of registration. Other fields are optional or cannot be set at
 registration, e.g., the `total_value` field can only be changed through the
-minting and burning of coins:
+minting and burning of coins.
 
 
 | Name                          | Type                                   | Mutable      | Description                                                                                                                             | Mandatory |
 | ----------------------------- | -------------------------------------- | ------------ | -----------------------------------------------------------------------------------------------------------------------                 | --------  |
-| to_xdx_exchange_rate          | FixedPoint32                           | Yes          | The _rough_ exchange rate from `C` to XDX. Used only for dual attestation threshold and mempool gas price normalization                 | Yes       |
+| to_xdx_exchange_rate          | FixedPoint32                           | Yes          | The _rough_ exchange rate from `C` to XDX. Used only for dual attestation threshold and mempool gas-price normalization.                | Yes       |
```
bob-wilson, line 137: Also, a general comment about this table: A number of the descriptions refer to "C" but there's nothing in the section to define what "C" is. It's pretty obvious from the context, but you could just as well omit it from the descriptions, or perhaps write "the currency" if necessary

dips/dip-20.md
```diff
@@ -171,79 +168,60 @@ determining exchange rates for value transfer between different currencies.
...
-coin for that currency off-chain. For example, for XUS the smallest
-off-chain denomination coin is one cent (1/100th of an XUS), so the
-`fractional_part` for XUS would be 100. If you wanted to translate from an
-on-chain value `v` to a number in XUS's currency units, you would multiply
-by 1,000,000 (XUS's `scaling_factor`) and then truncate all digits after the
-hundredths place.
-
-#### Mutability
-Immutable
+This immutable field defines the truncation that needs to be performed when transferring
```
bob-wilson, line 185: The phrasing of this needs some work. In many situations, it may be better to round instead of truncating, so we should be agnostic here. We should also avoid saying "needs to be performed" -- some applications may prefer to use smaller fractional units.

dips/dip-20.md
```diff
@@ -319,32 +297,11 @@ privilege](https://github.com/diem/dip/blob/master/dips/dip-2.md#permissions)
...
-```
-
-The `mint` function may only be called by an account with the
-[Treasury Compliance role](https://github.com/diem/dip/blob/master/dips/dip-2.md#roles)
-that has a `MintCapability<C>` resource published under it. Recall that only
-one mint capability resource is created for a currency when it is
-registered.
+Any account that has a [`MintCurrency<C>`](https://github.com/diem/dip/blob/master/dips/dip-2.md#permissions)
+privilege may mint, or create, a coin with a non-zero value.
+Recall that there is only one minting permission for each currency that can exist.
```
bob-wilson, line 304: Dangling phrase? Might be better as "Recall that only one minting permission can exist for each currency."?

dips/dip-20.md
```diff
@@ -507,7 +441,7 @@ operations are as follows, and may be called by anyone:
 ## <a name="currency_ops">Currency Operations</a>
 
 There are a number of different operations that may be performed to view and update the
-metadata for a given currency (viz. [Required information for registration of a currency
+metadata for a given currency (viz., [required information for registration of a currency
 and metadata](#metadata_spec)). They can be divided into "getters" (non-mutative),
```
bob-wilson, line 445: I'm not sure what is meant by "...and metadata" here. When you register a currency, you specify the metadata, so it seems like there is only one thing, not "currency and metadata"

dips/dip-20.md
```diff
@@ -635,12 +569,12 @@ state of the blockchain that should be kept. These are:
...
-When deciding the on-chain currency code for a currency,
-the module, as well as the type name chosen for the currency must match the on-chain currency
-code chosen for the currency. e.g., if you want to use currency code `"ABC"` for a
-currency, then the fully qualified Move type that is used for that currency on-chain
-_must_ be `0x1::ABC::ABC`.
-
-Because of this, and in addition:
-* An on-chain currency code must be a valid Move identifier; and
-* It can consist solely of uppercase alphanumeric characters.
+When deciding on the on-chain currency code for a currency, the module as
```
bob-wilson, line 572: Several places here refer to the "on-chain currency code". There's no distinction between on-chain and off-chain currency codes, is there? Each currency has a single currency code that must be used consistently throughout the system. I suggest dropping "on-chain" for at least most of these references.

bob-wilson (changes requested):

dips/dip-20.md
```diff
@@ -30,33 +30,35 @@ address](https://github.com/diem/dip/blob/master/dips/dip-2.md#roles) as
...
+Before moving forward, it's worth pinning down terminology. We'll use the
+following throughout the rest of this document:
+* A _**Diem currency**_ (or "currency") is a medium of exchange that Diem
+  has determined it will support on chain. Any Move type that has been
+  successfully registered using one of the [registration
+  mechanisms](#registration_and_metadata) defined in this DIP will be
+  considered to represent such a medium of exchange on chain.
+* An _**asset**_ in such an on-chain currency in Diem is a Move value that
+  represents a specific amount of spending power in the currency.
+  A _**coin**_ (or "Diem Coin") in such an on-chain currency is a specific
```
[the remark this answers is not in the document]
tzakian: Yes, this is correct.  

We want to talk about a Diem asset that is not also a Diem Coin when we are defining that the only representation of an asset on Diem is a Diem Coin. We could do this by saying "a Diem Coin" and a "Diem asset" are interchangeable, but IMO I think it's important to hammer home the point that "the _only_ representation of an asset on chain is as a Diem Coin."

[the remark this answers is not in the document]
bob-wilson: So you want to define "Diem asset" just so you can say that the only "Diem asset" is a "Diem Coin"? I get the point but it seems unnecessarily confusing. Just define "Diem Coin" and say that it is the only way to represent value in a currency.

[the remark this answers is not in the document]
tzakian: Good point! I've updated.

bob-wilson (approved):
It's a little hard to see exactly what you've changed from the previous version, but I think you've addressed my comments. I found a few more minor things that you might want to fix up, but I'll go ahead and approve this now.

dips/dip-20.md
```diff
@@ -30,33 +30,33 @@ address](https://github.com/diem/dip/blob/master/dips/dip-2.md#roles) as
...
-operations must be performed on a registered currency type, with the
-exception of registration.
+Before moving forward, it's worth pinning down terminology. We'll use the
+following throughout the rest of this document:
+* A _**Diem currency**_ (or "currency") is a medium of exchange that Diem
+  has determined it will support on chain. Any Move type that has been
+  successfully registered using one of the [registration
+  mechanisms](#registration_and_metadata) defined in this DIP will be
+  considered to represent such a medium of exchange on chain.
+* A _**coin**_ (or "Diem Coin") in such an on-chain currency is a specific
```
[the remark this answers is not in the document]
tzakian: Good spot, and very good point. I'll fix both :)

sblackshear (approved):

dahliamalkhi (commented):
Really happy to see this has shaped into a coherent and comprehensive overview of the way Assets are issued on the Diem blockchain.

I recommend moving this to "Last Call".

PR #139, opened 2021-02-12
Move DIP-20 from Draft to Last Call

tzakian:


bob-wilson (approved):
Dahlia suggested this and I concur.

dahliamalkhi (approved):

PR #141, opened 2021-03-06
Add dip-12 Coin Trade Flow

LuZhang-Lou:
Please see motivation and spec details in the doc itself.
Note this PR Intentionally only includes the on chain approach.


dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+```
+enum Metadata {
+  Undefined,
+  GeneralMetadata(GeneralMetadata),
+  TravelRuleMetadata(TravelRuleMetadata),
+  UnstructuredBytesMetadata(UnstructuredBytesMetadata),
+  RefundMetadata(RefundMetadat),
+  CoinTradeMetadata(CoinTradeMetadata),
+  /// and other types of metadatas
+}
```
davidiw, line 31: should really figure out a more reasonable way of communicating this without copying around this ever increasing blob....

davidiw (approved):
good as a first draft

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+created: 3/5/2021
+---
+# Summary and Motivation
+This document proposes a protocol standard for coin trade flows in the Diem Network. The flow is generic for coin purchase and coin sell transactions involving one or more Designated Dealers (DD). It supports a single trade settlement and batch settlement.
+
+A Blockchain coin trade is the process where an entity, for example a Virtual Asset Service Provider (VASP), transacts with a DD to purchase or sell coins. Beyond on-chain p2p transactions, a coin trade usually involves several stages ranging from trade initialization, fiat currency (or other types of cryptocurrency) debitted, to on-chain transfer. The first few steps may vary from entities to entities. However, we see the possibilities and benefits to provide a general specification for the on-chain phase, in pursuit of interoperability among Diem Blockchain participants.
+
+# Terminology
+* trade_id
+  * a unified trade identifier used over the course, a UTF-8 string, max length 255 bytes. It is agreed upon, by both entities, during the trade initialization and negotiation phase. It should be buyer-seller pair-wise unique.
```
davidiw, line 17: probably make a single line:
```
* ``trade_id`` - ...
```

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+dip: 12
+title: Blockchain Coin Trade Flow
+authors: Lu Zhang (@LuZhang-Lou)
+dip manager: David Wolinsky (@davidiw)
+status: Draft
+type: Informational
+created: 3/5/2021
+---
+# Summary and Motivation
+This document proposes a protocol standard for coin trade flows in the Diem Network. The flow is generic for coin purchase and coin sell transactions involving one or more Designated Dealers (DD). It supports a single trade settlement and batch settlement.
```
davidiw, line 11: "or more DDs?" how? considering the nature of this doc, I would just say:

This document proposes a standardized method for referring to off-chain purchases and sales of coins on-chain within Diem transaction metadata.

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+
+A Blockchain coin trade is the process where an entity, for example a Virtual Asset Service Provider (VASP), transacts with a DD to purchase or sell coins. Beyond on-chain p2p transactions, a coin trade usually involves several stages ranging from trade initialization, fiat currency (or other types of cryptocurrency) debitted, to on-chain transfer. The first few steps may vary from entities to entities. However, we see the possibilities and benefits to provide a general specification for the on-chain phase, in pursuit of interoperability among Diem Blockchain participants.
```
davidiw, line 13: always like bullets.

I might suggest we state:

The processes for off-chain transactions are typically unique to the parties involved, however, the on-chain transaction can likely be generalized to pointing to a unique identifier established during the coin trade.

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+
+/// Transaction metadata format for coin trades (purchases/sells)
+struct CoinTradeMetadataV0 {
+    /// A nonempty list of trade_ids this transaction wants to settle,
+    trade_ids: Vec<String>,
+}
+```
+
+
+`CoinTradeMetadata` holds a list of trade_ids that the coin transfer wants to settle. Refer to [Binary Canonical Serialization](https://github.com/diem/bcs#binary-canonical-serialization-bcs) for serialization/deserialization details, or use [Diem Client SDK](https://github.com/diem/client-sdks) if available.
```
davidiw, line 46: I think this is will or wants to, since the off-chain isn't dictated by this document.

davidiw: I don't think DIPs should refer to the SDKs.

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+metadata = diem_types.Metadata__RefundMetadata(
+    value=diem_types.RefundMetadata__RefundMetadataV0(
+        value=diem_types.RefundMetadataV0(
+            transaction_version=original_transaction_version,
+            reason=diem_types.RefundReason__OtherReason()
+        )
+    )
+)
+bcs_metadata = metadata.bcs_serialize()
+```
```
davidiw, line 67: I think this belongs in a specification / guide not a DIP. When I get back next week, we can figure out how we start crafting a spec, with samples, and dummy data

dips/dip-12.md
bmwill, line 17: > max length 255 bytes

If this is the case then the code that introduced this in the diem/diem repo should be updated to enforce this invariant.

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+
+/// List of supported transaction metadata format versions for coin trade transaction
+enum CoinTradeMetadata {
+    CoinTradeMetadataV0(CoinTradeMetadatav0),
+}
+
+/// Transaction metadata format for coin trades (purchases/sells)
+struct CoinTradeMetadataV0 {
+    /// A nonempty list of trade_ids this transaction wants to settle,
+    trade_ids: Vec<String>,
```
bmwill, line 41: Is there a limit to number of trade_ids (other than limited by txn size) that this should enforce?

dips/dip-12.md
bmwill: We should probably be much more explicit, for example instead of copying this list we should state what the serialized format looks like, e.g. what is the variant number that corresponds to this entry

dips/dip-12.md
bmwill: What is a DIP then? are DIPs not intended to be specifications?

dips/dip-12.md
LuZhang-Lou: I don't think there's hard limit given that this number should be small in practice (<< 100) . Are you suggesting to set a threshold here for  general guidance?

dips/dip-12.md
```diff
@@ -0,0 +1,67 @@
...
+# Specification
+We propose a new type of on-chain Metadata - `CoinTradeMetadata`.
+
+```
+enum Metadata {
+  Undefined,
+  GeneralMetadata(GeneralMetadata),
+  TravelRuleMetadata(TravelRuleMetadata),
+  UnstructuredBytesMetadata(UnstructuredBytesMetadata),
+  RefundMetadata(RefundMetadat),
```
prasincs, line 28: typo
```suggestion
  RefundMetadata(RefundMetadata),
```

dips/dip-12.md
prasincs: A general guidance would be good here. I'd expect most of the requests to have 1 tradeID? Trying to go over 100 likely points to some misunderstanding of how this field is expected to be used.

dips/dip-12.md
prasincs, line 17: > It should be buyer-seller pair-wise unique.

This could be expanded to read more clearly as "It should be unique for every trade between a buyer and a seller" 
Maybe there could be some guidance on example TradeIDs? like uint64 ids or UUIDs, etc?

dips/dip-12.md
LuZhang-Lou: I'm going to remove the part of "SDK"

dips/dip-12.md
LuZhang-Lou: this is recommendation/reference only, I don't want to introduce too many constraints/checks into the implementation. Let me rephrase this to be explicit. cc @CapCap

LuZhang-Lou: We don't specify the ids here because this type of IDs can vary a lot per DD. The hope is the buyer and sender can leverage the same id over the course and avoid adding another layer of translation

dips/dip-12.md
davidiw: I mistyped specification in the above. It should only really be in a guide unless we want embrace a single and consistent language for defining protocols.

dips/dip-12.md
LuZhang-Lou: IMO this should be decided by the involved buyer sand seller. If they want to settle 100 trades in a on-chain transfer, I don't see why we shouldn't allow them

dips/dip-12.md
```diff
@@ -0,0 +1,56 @@
...
+authors: Lu Zhang (@LuZhang-Lou)
+dip manager: David Wolinsky (@davidiw)
+status: Draft
+type: Informational
+created: 3/5/2021
+---
+# Summary and Motivation
+This document proposes a standardized method for coin trade flows in the Diem Network. The flow is generic for coin purchase and coin sell transactions. It supports a single trade settlement and batch settlement.
+
+
```
aching, line 13: nit: extra line

dips/dip-12.md
```diff
@@ -0,0 +1,56 @@
...
+A Blockchain coin trade is the process where an entity, for example a Virtual Asset Service Provider (VASP), transacts with a DD to purchase or sell coins. Beyond on-chain p2p transactions, a coin trade usually involves several stages, such as
+  1. trade initialization
+  2. fiat currency (or other types of cryptocurrency) debitted
```
aching, line 16:
```suggestion
  2. fiat currency (or other types of cryptocurrency) debited
```

dips/dip-12.md
```diff
@@ -0,0 +1,56 @@
...
+  3. on-chain transfer.
+ The first couple of steps are typically unique to the parties involved. However, we see the possibilities and benefits to generalize the on-chain phase, in pursuit of interoperability among Diem Blockchain participants.
+
+# Terminology
+* trade_id - a unified trade identifier used over the course, a UTF-8 string. The recommended max length is 255 bytes. It is agreed upon, by both entities, during the trade initialization and negotiation phase. It should be unique for every trade between a buyer and a seller.
```
aching, line 21: Why is a trade_id expected to be a UTF-8 string rather than a u64/u128 for instance?

dips/dip-12.md
```diff
@@ -0,0 +1,56 @@
...
+---
+# Summary and Motivation
+This document proposes a standardized method for coin trade flows in the Diem Network. The flow is generic for coin purchase and coin sell transactions. It supports a single trade settlement and batch settlement.
+
+
+A Blockchain coin trade is the process where an entity, for example a Virtual Asset Service Provider (VASP), transacts with a DD to purchase or sell coins. Beyond on-chain p2p transactions, a coin trade usually involves several stages, such as
+  1. trade initialization
+  2. fiat currency (or other types of cryptocurrency) debitted
+  3. on-chain transfer.
+ The first couple of steps are typically unique to the parties involved. However, we see the possibilities and benefits to generalize the on-chain phase, in pursuit of interoperability among Diem Blockchain participants.
```
davidiw, line 18: unique -> distinct

don't use first or third-persons in these (I, we)

dips/dip-12.md
davidiw: good call, best to say that at trade_id is an identifier defined by the two parties involved and is therefore represented by a UTF-8 string with a recommended maximum length of 255 bytes (or characters?).

dips/dip-12.md
```diff
@@ -0,0 +1,56 @@
...
+    CoinTradeMetadataV0(CoinTradeMetadatav0),
+}
+
+/// Transaction metadata format for coin trades (purchases/sells)
+struct CoinTradeMetadataV0 {
+    /// A nonempty list of trade_ids this transaction wants to settle
+    trade_ids: Vec<String>,
+}
+```
+
```
davidiw, line 48: remember to strip extra lines

davidiw (commented):
any idea why it is having issues with integration tests? note that the repo branch now is main instead of master

dips/dip-12.md
LuZhang-Lou: @aching we've seen non-numeric trade id from some partner. Usually this id is the one that is included in the trade confirmation docs, so if we make them numeric then it adds another layer of transformation for them.

dips/dip-12.md
```diff
@@ -0,0 +1,56 @@
...
+status: Draft
+type: Informational
+created: 3/5/2021
+---
+# Summary and Motivation
+This document proposes a standardized method for coin trade flows in the Diem Network. The flow is generic for coin purchase and coin sell transactions. It supports a single trade settlement and batch settlement.
+
+
+A Blockchain coin trade is the process where an entity, for example a Virtual Asset Service Provider (VASP), transacts with a DD to purchase or sell coins. Beyond on-chain p2p transactions, a coin trade usually involves several stages, such as
+  1. trade initialization
```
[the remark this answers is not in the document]
LuZhang-Lou: what do you mean ? 1 and 2 are both out-of-band.

davidiw (approved):

PR #142, opened 2021-03-06
[dip-1] add trade_ids in PaymentObject

LuZhang-Lou:
The motivation is to support passing trade_ids in Off Chain / Travel Rule protocol, such that mapping among trade_id, reference_id and on chain transfer can be mapped.

Please see details in https://docs.google.com/document/d/1DNaRzSjHGU7HD8xrEbU_cg3JY5apQI5VQcIpHM44EAI/edit?usp=sharing


davidiw (commented):
I personally don't believe this belongs here. We're already intending to make the ReferenceId command as separate from a KYC exchange. Similarly a trade id is intended to map to a reference id and should be established in another command. Secondly, this isn't clear to me if this is the standard way we want to be doing things in exchanges. We have a problem here with composition in that our existing TR is incompatible with other chains. This just further embraces that.

If we're comfortable that this is tech debt that we plan to clean up, I might look the other way :P

davidiw:
per conversation, please move this to an issue -- we could consider this another dip though too or at least leave it in a draft state

PR #148, opened 2021-03-21
DIP-1 becomes purely about off-chain

davidiw:
* Move Travel rule into an independent DIP-13
* Add an optional result field to responses
* Improve statements around cid on responses
* Use must and instead of should around UUID requirements
* Removed example code -- should be in a guide not in a spec
* Introduce a PingCommand, optional
* Make author's list inclusive

Action items:
* DIP-13 must be created, approved, and landed before landing this
* I've added a list of authors, if you don't feel your name belongs there, let me know, but I'm a big fan of being inclusive
* @xli, I've noticed that the wording around cid in response makes it quite clear that it is expected to be there unless the request is unparseable, but I remember a conversation where maybe that thought wasn't consistent either in DRW or SDK or both

### DIFF
--- a/dips/dip-1.mdx
+++ b/dips/dip-1.mdx
@@ -31,11 +31,15 @@ The two parties that participate in the off-chain protocol communicate through H
 
 # Specification
 ---
-## On-Chain Account Setup {#on-chain-account-setup}
+## On-Chain Account Setup
 
 A VASP’s on-chain account must be created and set up in accordance with the standard VASP account creation process.
 
-Of particular note for off-chain APIs is the compliance key and base url, [DualAttestation::Credential](https://github.com/diem/diem/blob/release-1.1/language/stdlib/modules/DualAttestation.move)’s `compliance_public_key` and `base_url`. The compliance key value is the Ed25519 key with which a VASP signs travel rule statements which are verified on-chain and all off-chain requests and responses. During account setup, a VASP should set their compliance key and base url which will be stored under the Parent VASP account, see [DIP-2](https://github.com/diem/dip/blob/master/dips/dip-2.md) for more information on on-chain roles.
+Of particular note for off-chain APIs is the compliance key and base url, [DualAttestation::Credential](https://github.com/diem/diem/blob/release-1.1/language/stdlib/modules/DualAttestation.move)’s `compliance_public_key` and `base_url`. During account setup, a VASP should set their compliance key and base url which will be stored under the Parent VASP account, see [DIP-2](https://github.com/diem/dip/blob/master/dips/dip-2.md) for more information on on-chain roles.
+
+The compliance key value is the Ed25519 key with which a VASP signs travel rule statements which are verified on-chain and all off-chain requests and responses.
+
+The `base_url` frameowrk does not support multiplexing and should not be shared across different providers and accounts.
 
 ## HTTP End-point
 
@@ -55,9 +59,8 @@ The HTTP request sender must use the compliance key of the VASP account linked w
 All HTTP responses must contain:
 
 * A header `X-REQUEST-ID` copied from the HTTP request.
-* A header `X-DIEM-ADDRESS`, the value should be a hex-encoded diem onchain account address bytes that can be used for looking up the compliance key that can decode the response body JWS message.
 
-The HTTP response sender must use the compliance key of the VASP account linked with the responder's address to sign the response JWS body. The response receiver uses this address to find the appropriate compliance key to verify the signed request. For example: VASP A transfers funds to VASP B. The HTTP response from B to A contains `X-DIEM-ADDRESS` as VASP B's address.
+The HTTP response sender must use the compliance key of the VASP account linked with the responder's address to sign the response JWS body.
 
 ## Payloads
 
@@ -191,7 +194,6 @@ The following sections list all error codes for various validations when process
 * Could not find the compliance key of the onchain account found by the `X-REQUEST-SENDER-ADDRESS` header value.
 * The compliance key found from the onchain account by `X-REQUEST-SENDER-ADDRESS` is not a valid ED25519 public key.
 * `X-REQUEST-ID` is not in valid UUID format.
-* `X-DIEM-ADDRESS` is not valid.
 
 `missing_http_header`: missing HTTP header `X-REQUEST-ID` or `X-REQUEST-SENDER-ADDRESS`.
 
@@ -301,7 +303,6 @@ V3 changes:
 * Add an optional result field to responses
 * Improve statements around cid on responses
 * Use must and instead of should around UUID requirements
-* Add X-DIEM-ADDRESS as required -- this seems ... how did this get
   added to the specification as required?
 * Removed example code -- should be in a guide not in a spec
 * Introduce a PingCommand
--- a/dips/dip-1.mdx
+++ b/dips/dip-1.mdx
@@ -1,7 +1,7 @@
 ---
 dip: 1
 title: Off-chain API
-authors: Kevin Hurley (@kphfb), Dmitry Pimenov, George Danezis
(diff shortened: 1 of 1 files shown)


dips/dip-1.mdx
```diff
@@ -11,141 +11,77 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-
-If VASP A fails to receive a response from VASP B, or in the case of specific protocol failure codes, it must resend the request at some cadence until a response is received to ensure consistency.
-
-<img alt="Command Exchange" src={useBaseUrl('img/command_exchange.png')} />
-
-As mentioned in [Terminology](#terminology), every Object type contains a `reference_id` field which is a unique reference ID for the Object. At each state of an Object, only one VASP is allowed to mutate it (or none if the Object is in a terminal state). If the other VASP tries to mutate the object, the command will be rejected. This prevents concurrent attempts to update Objects that could lead to inconsistencies.
-
+The HTTP response sender must use the compliance key of the VASP account linked with the responder's address to sign the response JWS body. The response receiver uses this address to find the appropriate compliance key to verify the signed request. For example: VASP A transfers funds to VASP B. The HTTP response from B to A contains `X-DIEM-ADDRESS` as VASP B's address.
```
xli, line 60: The request sender already knows a receiver diem address, should it expect same address here? Or different x-diem-address is ok? I will assume they are same and validate the x-diem-address is same (same parent vasp or dd account address)

dips/dip-1.mdx
```diff
@@ -175,17 +110,19 @@ All requests between VASPs are structured as `CommandRequestObject`s.
...
 | Field 	     | Type 	| Required? 	| Description 	|
 |-------	     |------	|-----------	|-------------	|
 | _ObjectType    | str      | Y             | The fixed string `CommandResponseObject`. |
 | status         | str      | Y             | Either `success` or `failure`. |
-| error          | [OffChainErrorObject](#offchainerrorobject) | N | Details of the error when status == "failure" |
-| cid            | str      | N | The Command identifier to which this is a response. Should be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included and should match the 'cid' of the CommandRequestObject. |
-
+| error          | [OffChainErrorObject](#offchainerrorobject) | N | Details of the error when status == "failure". |
+| result         | Object | N | An optional JSON object that may be defined when status == "success". |
+| cid            | str      | N | The Command identifier to which this is a response. Must be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included and must match the 'cid' of the CommandRequestObject. This field must be set unless the request to which this is responding is unparseable. |
```
xli, line 123: Other cases:
1. No cid in the request 
2. cid in the request is not a valid uuid
3. Blob is valid json, but there is a field type does not match (for example: command_type value is a int), causes the deserialization failure. You probably can still parse out the cid by not deserializing json blob in this case, but need call out and handle it probably.

dips/dip-1.mdx
```diff
@@ -11,141 +11,77 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-**Shared Object**: All Objects contained within a Command are logically shared, meaning that each VASP has a copy of the Object and may create a new Command to modify it. For example, during the lifecycle of a `KycDataObject`, both VASPs will add information to it. The protocol does not support simultaneous updates to Objects. Rather, VASPs must take turns updating the Object - with turns specified as per the [command sequence](#paymentobject-protocol-command-sequence).
+## HTTP End-point
 
-**Reference ID**: Every Object type contains a `reference_id` field which is a unique reference ID for the Object. The reference ID is always specified by the payment initiator VASP (the VASP which originally created the Object). This value should be globally unique. It is recommended to use a 128 bit long UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included.
+Each VASP exposes an HTTPS POST end point at `https://&lt;hostname>:&lt;port>/&lt;protocol_version>/command`. The protocol_version is v2 for this iteration of the off-chain APIs.
```
xli, line 42: Current version need bump up to "v3"

bmwill: I thought the intent of some of these changes would be to make it be done in a compatible way so that you don't need to bump the version number. If we're bumping the version number then we can and should make a lot more changes.

davidiw: Why are we bumping it to v3? This doesn't contain any breaking changes with the current protocol. Which is why I'm not sure why we included new mandatory fields in the specification doc ;)

dips/dip-1.mdx
davidiw: I think all of these imply that the request is unparseable

dips/dip-1.mdx
davidiw: I didn't introduce this, who did introduce X-DIEM-ADDRESS? I am very much in favor of deleting it or making this a could / optional field.

davidiw: Discussed offline, there are a dozen different ways to handle this than introducing a new field, so I will strip this out.

davidiw: deleted and updated in the current update

davidiw:
https://github.com/diem/dip/issues/152

pdhamdhere:
DIP-1 is already in **Final** stage. Please note  partners have implemented or in the process of implementing DIP-1 as it stands today. Changing the Spec which is already final can not only create confusion but may introduce compatibility issues in production. Would the right [process](https://dip.diem.com/overview/#the-dip-process) be to start new DIP and then deprecate DIP-1?

davidiw:
> DIP-1 is already in **Final** stage. Please note partners have implemented or in the process of implementing DIP-1 as it stands today. Changing the Spec which is already final can not only create confusion but may introduce compatibility issues in production. Would the right [process](https://dip.diem.com/overview/#the-dip-process) be to start new DIP and then deprecate DIP-1?

What makes you feel like this is an incompatible update to the dip? It's largely improving the readability and improving the organization. If it breaks anything or creates any incompatibilities, then the update has failed it's purpose.

With respect to a new dip, the intent here being a forward compatible update, we've agreed to using versioning. If we were to introduce a completely different off chain format, say ethereum whisper, it would be a different dip.

aching:
While I do appreciate that incompatible updates to a protocol are not as disruptive, I think this DIP has gone substantive changes over time  and we should be following the process of DIP deprecation rather than updating it.  Also as this is an optional standard, it makes sense to keep it in a draft stage until we have sufficient confidence to more to Accepted/Final (this would avoid multiple DIP deprecation/updates).

From https://dip.diem.com/overview
Deprecated – The DIP Manager or Maintainers may mark a DIP Deprecated if it is superseded by a later proposal or otherwise becomes irrelevant.
Final – The Final state of a DIP means the necessary implementations of the DIP are complete and deployed to the codebase. This DIP represents the current state-of-the-art. A Final DIP should only be updated to correct errata. (edited).

bmwill:
I think this highlights a few issues with the DIP process as a whole. Specifically that DIPs cannot be looked at as a cohesive Specification. In particular it might be more ideal to have a single offchain Specification (that includes everything) and instead use the DIP process as a way to propose and ratify changes to said specification.

dips/dip-1.mdx
xli: What you mean unparsable? The request json string is parsed successfully for all above cases.

dips/dip-1.mdx
xli: I think bump up version for the change added.
I feel confusing when doc has v3, but the protocol version is still v2. 

But up to you.

davidiw:
At this point, the plan is to take this content and leverage it in a spec instead of updating the DIP. Once the spec is fully written, we should eliminate referring to DIPs from our nomenclature and instead refer to the spec.

PR #151, opened 2021-03-22
Update DIP-10

sunmilee:
Updating DIP-10 to reflect the current PayAddress MVP spec. Changes include:
- Include using CSV file for domain distribution 
- Include Reference ID exchange
- Remove UserObject and ProfileURI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@
 .docusaurus
 .cache-loader
 DIPMetadata.json
-.idea
 
 # Misc
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .docusaurus
 .cache-loader
 DIPMetadata.json
+.idea
 
 # Misc
 .DS_Store
--- a/dips/dip-10.md
+++ b/dips/dip-10.md
@@ -1,8 +1,7 @@
 ---
 dip: 10
-title: Diem ID Spec
-authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb), Sunmi Lee (@sunmilee)
-, David Wolinsky (@davidiw)
+title: Pay Address Spec
+authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb), Sunmi Lee (@sunmilee), David Wolinsky (@davidiw)
 status: Draft
 type: Informational
 created: 11/03/2020
@@ -12,15 +11,15 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 
 # Summary
 
-This DIP describes Diem ID - the human-readable identifier for user accounts.   
+This DIP describes Pay Address - the human-readable identifier for user accounts.   
 
 # Motivation
-Diem ID provides a convenient way to identify a user's account within a VASP.
-* For peer-2-peer payments, Diem ID allows a user to provide a human-readable ‘identifier’ to the sender. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
-  * Diem ID provides better privacy compared to using sub-addresses, because PayAddress is not shared on the chain and will always use a single-use identifier for the on-chain ID
-  * From a user's perspective, Diem ID is a persistent identifier.  But from the perspective of the chain, Diem IDs do not exist and instead transactions are have a single-use ID to identify them. By contrast, from the viewpoint of both a user and the chain, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy, but appears on-chain and could thus be used to link payments if it isn't recycled upon each usage.
+Pay Address provides a convenient way to identify a user's account within a VASP.
+* For peer-2-peer payments, Pay Address allows a user to provide a human-readable ‘identifier’ to the sender. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Pay Address once. In this case Pay Address plays the role of an email address for payments. Benefits for this use case:
+  * Pay Address provides better privacy compared to using sub-addresses, because PayAddress is not shared on the chain and will always use a single-use identifier for the on-chain ID
+  * From a user's perspective, Pay Address is a persistent identifier.  But from the perspective of the chain, Pay Addresss do not exist and instead transactions are have a single-use ID to identify them. By contrast, from the viewpoint of both a user and the chain, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy, but appears on-chain and could thus be used to link payments if it isn't recycled upon each usage.
 
-# Diem ID Format
+# Pay Address Format
 Diem defines globally unique identifiers to facilitate transacting on the Diem Payment Network. As Diem stores no PII on-chain, these identifiers are exchanged during the off-chain process between two VASPs to identify a distinct source and destinations within their respective VASPs in order to produce a reference_id that can be stored on-chain. The format for the identifiers follows:
 `[user_identifier]@[vasp_domain_identifier]`
 
@@ -37,14 +36,14 @@ Example: `alice@avasp`
 
 
 # Naming Service (Off-Chain Lookup Service)
-The Diem ID specifies that VASPs can receive payments at a vasp_domain_identifier. The intent is to eventually store these values on-chain. As an initial effort into this space, the Diem Association stores a file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
+The Pay Address specifies that VASPs can receive payments at a vasp_domain_identifier. The intent is to eventually store these values on-chain. As an initial effort into this space, the Diem Association stores a file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
 ```
 avasp, 0xf72589b71ff4f8d139674a3f7369c69b
 bvasp, 0xc5ab123458df0003415689adbb47326d
 ```
 
 # On-chain data
-## Diem ID Domain
+## Pay Address Domain
 
 ```
 resource struct DiemIdDomains {
@@ -56,21 +55,21 @@ struct DiemIdDomain {
 }
 ```
 * Fields definition:
-   * `domain` - name of a Diem ID domain 
+   * `domain` - name of a Pay Address domain 
 * The `DiemIdDomains` resource can only be published into an account with `Role == PARENT_VASP_ROLE_ID`.
-* The `DiemIdDomains` contains a list of `DiemIdDomain` structs that are associated with a VASP. As such, it is possible to register more than one Diem ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
+* The `DiemIdDomains` contains a list of `DiemIdDomain` structs that are associated with a VASP. As such, it is possible to register more than one Pay Address Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
 * Only special Treasury Compliance account (address `0xB1E55ED`) can manipulate DiemIdDomains resource:
   * Only TC account can create and publish `DiemIdDomains` resource
   * Only TC account can add, remove or update an `DiemIdDomain` within `DiemIdDomains` resource
-* In order to register a Diem Id domain, a VASP needs to submit a request to Diem Association. LA will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem Id Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
-* The Diem ID domain can be created as part of creating a `ParentVASP` account.
-* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
+* In order to register a Pay Address domain, a VASP needs to submit a request to Diem Association. LA will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Pay Address Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
+* The Pay Address domain can be created as part of creating a `ParentVASP` account.
+* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Pay Addresss and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
 
-## Diem ID Domain Events
+## Pay Address Domain Events
 
-The Move contract that manages Diem ID domains must emit an event every time Diem ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Diem ID domains.
+The Move contract that manages Pay Address domains must emit an event every time Pay Address domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Pay Address domains.
 
-While Diem ID domains are published into VASP accounts, Diem ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
+While Pay Address domains are published into VASP accounts, Pay Address domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
 
 To store events, `DiemIdDomainManager` resource is published under the Treasure Compliance account(address `0xB1E55ED`).
 
@@ -97,7 +96,7 @@ struct DiemIdDomainEvent {
 The PaymentMetadata transactions require a ReferenceID in order to submit a transaction to the chain. PaymentMetadata transactions reveal nothing distinctly about either the sender or recipient and do not create a linkability between the sender or receiver across payments. Hence, the ReferenceID must be established off-chain and this protocol defines one such way to do so leveraging the PaymentAddresses.
 
 ## ReferenceID Command
-A VASP intending to send a payment from one VASP to another when leveraging Diem IDs constructs a ReferenceIDCommand submits that to the recipient side. If the recipient knows the recipient PaymentAddress and can potentially accept payments, it returns a success. In the case it cannot, it returns an error indicating the reason.
+A VASP intending to send a payment from one VASP to another when leveraging Pay Addresss constructs a ReferenceIDCommand submits that to the recipient side. If the recipient knows the recipient PaymentAddress and can potentially accept payments, it returns a success. In the case it cannot, it returns an error indicating the reason.
 
 In the case that the amount to be sent would exceed the limit of the travel rule, the sending party should follow this exchange with a PaymentCommand using the same reference_id and specify the sending and receiving subaddresses as all 0.
 
@@ -132,7 +131,7 @@ The format of the success response is:
 
 
 # End-to-End Experience
-Below is an example of using Diem ID domain for transferring money from one user to another. 
+Below is an example of using Pay Address domain for transferring money from one user to another. 
 
 ## Prerequisite:
 * VASPs get approval from association (via some offline process) on domain name
@@ -145,9 +144,9 @@ novi, 0xc5ab123458df0003415689adbb47326d
 
 ## User Story: 
 * Bob wants to receive funds from Alice
-* Bob registers a Diem ID at his VASP (bvasp): bob@bvasp
-* Bob shares either this Diem ID, bob@bvasp, or a PaymentURI, diem://bob@bvasp with Alice
+* Bob registers a Pay Address at his VASP (bvasp): bob@bvasp
+* Bob shares either this Pay Address, bob@bvasp, or a PaymentURI, diem://bob@bvasp with Alice
 * Alice logs into her VASP (avasp), enters Bob’s identifier, an amount to pay, and submits payment
 * Alice’s VASP contacts Bob’s VASP’s off-chain API with the sender identifier (alice@avasp) and requests a reference_id, rb1
 * Alice’s VASP constructs a transaction with the specified amount and the reference_id (rb1) and submits it to the diem network
-* Bob’s VASP receives a transaction with metadata containing rb1, deposits the amount in his account, and attaches the relevant metadata (e.g., Alice’s Diem ID) so Bob can confirm the transaction
+* Bob’s VASP receives a transaction with metadata containing rb1, deposits the amount in his account, and attaches the relevant metadata (e.g., Alice’s Pay Address) so Bob can confirm the transaction
--- a/dips/dip-10.md
+++ b/dips/dip-10.md
@@ -14,10 +14,9 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 This DIP describes Diem ID - the human-readable identifier for user accounts.   
 
 # Motivation
-Diem ID provides a convenient method for identifying users within a VASP.
-* For peer-2-peer payments, Diem ID allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
-  * Privacy: There are concerns around leakage of PII when using subaddresses. Diem ID provides better privacy compared to sub-addresses, because unlike sub-addresses, Diem ID is intended to be used in tandem with an off-chain protocol. Hence, Diem ID does not need to expose PII to the chain 
-  * Persistent Identifiers: Currently there is a lack of persistent user identifiers in the Diem ecosystem. while subaddresses need to be refreshed for each payment, Diem ID is a persistent identifier from a user's perspective. From the perspective of the chain, Diem IDs does not exist.
+Diem ID provides a convenient method for identifying users within a VASP. For peer-2-peer payments, Diem ID allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
+* Privacy: There are concerns around leakage of personally identifiable information (PII) when using subaddresses. Diem ID provides better privacy compared to sub-addresses, because unlike sub-addresses, Diem ID is intended to be used in tandem with an off-chain protocol. Hence, Diem ID eliminates the need to expose PII on the chain 
+* Persistent Identifiers: Currently there is a lack of persistent user identifiers in the Diem ecosystem. while subaddresses need to be refreshed for each payment, Diem ID is a persistent identifier from a user's perspective. From the perspective of the chain, Diem IDs does not exist.
 
 # End-to-End Experience
 Below is an example of using Diem ID domain for transferring money from one user to another. 
@@ -33,11 +32,11 @@ bvasp, 0xc5ab123458df0003415689adbb47326d
 * Bob wants to receive funds from Alice
 * Bob registers a Diem ID at VASP B: `bob@bvasp`
 * Alice's Diem ID is registered at VASP A: `alice@avasp`
-* Bob shares either this Diem ID, `bob@bvasp`, or a PaymentURI, `diem://bob@bvasp` with Alice
+* Bob shares either this Diem ID, `bob@bvasp` with Alice
 * Alice logs into VASP A, enters Bob’s identifier, an amount to pay, and submits payment
 * Alice’s VASP (VASP A) contacts Bob’s VASP’s (VASP B) off-chain API with the sender identifier `alice@avasp` and requests a reference_id, `rb1`
 * VASP A constructs a transaction with the specified amount and the reference_id `rb` and submits it to the diem network
-* VASP B receives a transaction with metadata containing `rb1`, deposits the amount in his account, and attaches the relevant metadata (e.g., Alice’s Diem ID) to the transaction so Bob can confirm the transaction
+* VASP B receives a transaction with metadata containing `rb`, deposits the amount in Bob's account, and attaches the relevant metadata (e.g., Alice’s Diem ID) to the transaction so Bob can confirm the transaction
 
 
 # Diem ID Format
@@ -78,19 +77,18 @@ struct DiemIdDomain {
 * Fields definition:
    * `domain` - name of a Diem ID domain 
 * The `DiemIdDomains` resource can only be published into an account with `Role == PARENT_VASP_ROLE_ID`.
-* The `DiemIdDomains` contains a list of `DiemIdDomain` structs that are associated with a VASP. As such, it is possible to register more than one Diem ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
+* The `DiemIdDomains` contains a list of `DiemIdDomain` structs that are associated with a VASP. As such, it is possible to register more than one Diem ID Domain for a given VASP
 * Only special Treasury Compliance account (address `0xB1E55ED`) can manipulate DiemIdDomains resource:
   * Only TC account can create and publish `DiemIdDomains` resource
   * Only TC account can add, remove or update an `DiemIdDomain` within `DiemIdDomains` resource
-* In order to register a Diem ID domain, a VASP needs to submit a request to Diem Association. LA will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem ID Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
-* The Diem ID domain can be created as part of creating a `ParentVASP` account.
-* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
+* In order to register a Diem ID domain, a VASP needs to submit a request to Diem Association. Diem Association will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem ID Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
+* `DiemIDDomains` resource will be created as part of creating a `ParentVASP` account resource, and existing `ParentVASP` accounts without `DiemIDDomains` will have the resource instantiated by the DiemRoot account.
 
 ## Diem ID Domain Events
 
 The Move contract that manages Diem ID domains must emit an event every time Diem ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Diem ID domains.
-
-While Diem ID domains are published into VASP accounts, Diem ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
+An application can be built to listen for these events to construct a mapping of Diem ID domains to VASP accounts for lookup of onchain addresses given a Diem ID domain.
+While Diem ID domains are published into VASP account resource, Diem ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow a single event stream.
 
 To store events, `DiemIdDomainManager` resource is published under the Treasure Compliance account(address `0xB1E55ED`).
 
@@ -116,7 +114,6 @@ struct DiemIdDomainEvent {
 # ReferenceID Exchange
 The PaymentMetadata transactions require a ReferenceID in order to submit a transaction to the chain. PaymentMetadata transactions reveal nothing distinctly about either the sender or recipient and do not create a linkability between the sender or receiver across payments. Hence, the ReferenceID must be established off-chain and this protocol defines one such way to do so leveraging Diem ID.
 
-## ReferenceID Command
 A VASP intending to send a payment from one VASP to another when leveraging Diem IDs constructs a ReferenceIDCommand submits that to the recipient side. If the recipient knows the recipient Diem ID and can potentially accept payments, it returns a success. In the case it cannot, it returns an error indicating the reason.
 
 In the case that the amount to be sent would exceed the limit of the travel rule, the sending party should follow this exchange with a PaymentCommand using the same reference_id and specify the sending and receiving subaddresses as all 0.
--- a/dips/dip-10.md
+++ b/dips/dip-10.md
@@ -14,9 +14,9 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 This DIP describes Diem ID - the human-readable identifier for user accounts.   
 
 # Motivation
-Diem ID provides a convenient method for identifying users within a VASP. For peer-2-peer payments, Diem ID allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
-* Privacy: There are concerns around leakage of personally identifiable information (PII) when using subaddresses. Diem ID provides better privacy compared to sub-addresses, because unlike sub-addresses, Diem ID is intended to be used in tandem with an off-chain protocol. Hence, Diem ID eliminates the need to expose PII on the chain 
-* Persistent Identifiers: Currently there is a lack of persistent user identifiers in the Diem ecosystem. while subaddresses need to be refreshed for each payment, Diem ID is a persistent identifier from a user's perspective. From the perspective of the chain, Diem IDs does not exist.
+Diem ID provides a convenient method for identifying users within a VASP. Diem ID allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments, and plays the role of an email address for payments. The benefits of using a Diem ID are:
+* Privacy: Diem ID needs to be used in tandem with an off-chain protocol between VASPs in order exchange information about the end user corresponding to the Diem ID. Hence, Diem ID eliminates the need to directly include user information in a transaction and expose personally identifiable information (PII) on the chain 
+* Persistent Identifiers: Currently there is a lack of persistent user identifiers in the Diem ecosystem. Diem ID is a persistent identifier from a user's perspective. From the perspective of the chain, Diem IDs do not exist.
 
 # End-to-End Experience
 Below is an example of using Diem ID domain for transferring money from one user to another. 
@@ -32,11 +32,11 @@ bvasp, 0xc5ab123458df0003415689adbb47326d
 * Bob wants to receive funds from Alice
 * Bob registers a Diem ID at VASP B: `bob@bvasp`
 * Alice's Diem ID is registered at VASP A: `alice@avasp`
-* Bob shares either this Diem ID, `bob@bvasp` with Alice
+* Bob shares the Diem ID, `bob@bvasp` with Alice
 * Alice logs into VASP A, enters Bob’s identifier, an amount to pay, and submits payment
 * Alice’s VASP (VASP A) contacts Bob’s VASP’s (VASP B) off-chain API with the sender identifier `alice@avasp` and requests a reference_id, `rb1`
-* VASP A constructs a transaction with the specified amount and the reference_id `rb` and submits it to the diem network
-* VASP B receives a transaction with metadata containing `rb`, deposits the amount in Bob's account, and attaches the relevant metadata (e.g., Alice’s Diem ID) to the transaction so Bob can confirm the transaction
+* VASP A constructs a transaction with the specified amount and the reference_id `rb` and submits it to the Diem network
+* VASP B receives a transaction with metadata containing `rb`, deposits the amount in Bob's account, and attaches the relevant metadata (e.g., Alice’s Diem ID) to the internally stored transaction so Bob can confirm the transaction
 
 
 # Diem ID Format
@@ -45,18 +45,18 @@ Diem defines globally unique identifiers to facilitate transacting on the Diem P
 
 Example: `alice@avasp`
 
-* `user_identifier` is a unique identifier meaningful only to a VASP that represents either the source or destination of a transaction. In practice, they are likely to be re-used and should be capable of both being the sender and recipient in a transaction. Their specific lifetime is undefined but should ideally be meaningful to the owner of the identifier. Specification:
+* `user_identifier` is a reusable identifier that represents either the source or destination end-user of a payment transaction. It is unique to per user at VASP level. Specification:
   * Case insensitive
-  * Valid regular expression: [a-zA-Z0-9]*
+  * Valid regular expression: [a-zA-Z0-9]+
   * Maximum length: 64 characters
-* `vasp_domain_identifier` is a well-known string that can be mapped to a VASP. Specification:
+* `vasp_domain_identifier` is a unique string that is mapped to a VASP. Specification:
   * Case insensitive
-  * Valid regular expression: [a-zA-Z0-9]*
+  * Valid regular expression: [a-zA-Z0-9]+
   * Maximum length: 63 characters
 
 
 # Naming Service (Off-Chain Lookup Service)
-The Diem ID specifies that VASPs can receive payments at a vasp_domain_identifier. The intent is to eventually store these values on-chain. As an initial effort into this space, the Diem Association stores a file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
+The intent is to eventually store Diem ID domains on-chain; however, the Diem Association will initially store and distribute a CSV file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
 ```
 avasp, 0xf72589b71ff4f8d139674a3f7369c69b
 bvasp, 0xc5ab123458df0003415689adbb47326d
@@ -81,8 +81,8 @@ struct DiemIdDomain {
 * Only special Treasury Compliance account (address `0xB1E55ED`) can manipulate DiemIdDomains resource:
   * Only TC account can create and publish `DiemIdDomains` resource
   * Only TC account can add, remove or update an `DiemIdDomain` within `DiemIdDomains` resource
+* `DiemIDDomains` resource will be created in an empty state as part of creating a `ParentVASP` account resource, and existing `ParentVASP` accounts without `DiemIDDomains` will have the resource instantiated by the DiemRoot account.
 * In order to register a Diem ID domain, a VASP needs to submit a request to Diem Association. Diem Association will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem ID Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
-* `DiemIDDomains` resource will be created as part of creating a `ParentVASP` account resource, and existing `ParentVASP` accounts without `DiemIDDomains` will have the resource instantiated by the DiemRoot account.
 
 ## Diem ID Domain Events
 
@@ -146,3 +146,8 @@ The format of the success response is:
     "cid": "12ce83f6-6d18-0d6e-08b6-c00fdbbf085a",
 }
 ```
+
+# Diem ID or Subaddress?
+[DIP-5](https://github.com/diem/dip/blob/master/dips/dip-5.md) defines subaddresses as a way of defining a user address. Between subaddress and Diem ID, the recommended way of representing a user address is to use a Diem ID, as Diem ID provides clear benefits over subaddresses:  
+* Subaddresses are ideally one-time use addresses, and need to be refreshed for each payment. However, a Diem ID can be shared once and will be a persistent identifier from a user's perspective.
+* There are concerns around leakage of PII when using subaddresses, because the subaddress has to be included in the on-chain transaction metadata. Diem ID provides better privacy, because Diem ID is exchanged off-chain.
--- a/dips/dip-10.md
+++ b/dips/dip-10.md
@@ -1,7 +1,7 @@
 ---
 dip: 10
 title: Diem ID Spec
-authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb), Sunmi Lee (@sunmilee), David Wolinsky (@davidiw)
+authors: Sunmi Lee (@sunmilee), David Wolinsky (@davidiw), Andrey Chursin(@andll), Kevin Hurley (@kphfb)
 status: Draft
 type: Informational
 created: 11/03/2020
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@
 .docusaurus
 .cache-loader
 DIPMetadata.json
+.idea
```
davidiw, line 12: I think the intent here is that this should be in your local ignore and not the shared ignore for this repo. I don't use idea, I use vim, so if everyone included special files for their favorite editor, this would become unmanageable and unclear what is doing what.

dips/dip-10.md
```diff
@@ -1,7 +1,8 @@
 ---
 dip: 10
 title: Diem ID Spec
-authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb)
+authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb), Sunmi Lee (@sunmilee)
+, David Wolinsky (@davidiw)
```
davidiw, line 5: weird lining here, why isn't this just one line?

dips/dip-10.md
```diff
@@ -13,47 +14,37 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-  * The Diem ID resolution process can facilitate the exchange of rich information(e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
-  * Diem ID provides better privacy compared to using sub-addresses, because Diem ID is not shared on the chain and will always use a single-use identifier for the on-chain ID
+Diem ID provides a convenient way to identify a user's account within a VASP.
+* For peer-2-peer payments, Diem ID allows a user to provide a human-readable ‘identifier’ to the sender. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
```
davidiw, line 19: thinking something like: allows users to exchange human-readable identifiers as both the sender and recipient of payments peer-to-peer payments

dips/dip-10.md
```diff
@@ -13,47 +14,37 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-In addition to describing the format of the Diem ID, this DIP also describes the off-chain protocol for the peer-2-peer payment with Diem ID.
-
 # Motivation
-Diem ID provides a convenient way to identify a user's account within a VASP. There are two main use cases for Diem ID:
-* For peer-2-peer payments, Diem ID allows a user to provide a human-readable ‘identifier’ to the sender. Basically, instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that needs to be refreshed for each payment, ideally), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
-  * The Diem ID resolution process can facilitate the exchange of rich information(e.g. nicknames, profile picture, etc.) about both the sender and receiver of the payment
-  * Diem ID provides better privacy compared to using sub-addresses, because Diem ID is not shared on the chain and will always use a single-use identifier for the on-chain ID
+Diem ID provides a convenient way to identify a user's account within a VASP.
```
davidiw, line 18: convenient method for identifying users within a VASP

dips/dip-10.md
```diff
@@ -13,47 +14,37 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* For peer-2-peer payments, Diem ID allows a user to provide a human-readable ‘identifier’ to the sender. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
+  * Diem ID provides better privacy compared to using sub-addresses, because PayAddress is not shared on the chain and will always use a single-use identifier for the on-chain ID
```
davidiw, line 20: PayAddress or DiemID?

Note DiemID versus Diem ID, which is it :P

davidiw: I would rephrase this in that DiemID is intended to be used in tandem with an off-chain protocol, hence unlike sub-addresses DiemID need not expose personally identifiable information to the chain.

dips/dip-10.md
```diff
@@ -13,47 +14,37 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
   * From a user's perspective, Diem ID is a persistent identifier.  But from the perspective of the chain, Diem IDs do not exist and instead transactions are have a single-use ID to identify them. By contrast, from the viewpoint of both a user and the chain, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy, but appears on-chain and could thus be used to link payments if it isn't recycled upon each usage.
```
davidiw, line 21: this is a bit redundant

I might suggest that DiemID attempts to solve the current challenges with P2P payments on the Diem blockchain:
* Lack of persistent user identifiers
* Concerns around sub-addresses and leakage of pii

Then maybe a sentence or two for each point

dips/dip-10.md
```diff
@@ -101,130 +92,62 @@ struct DiemIdDomainEvent {
...
-  * Ex.: *https://novi.com/pictures/alice.png*
-* **Data schema.** If URI starts with data:, wallet must interpret the URI as a [Data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) that contains profile picture encoded.
-  * Ex.: *data:image/gif;base64,AAAAAAAA*
-* **Unencrypted http** schema must not be used to share profile pictures.
+coinbase, 0xf72589b71ff4f8d139674a3f7369c69b
+novi, 0xc5ab123458df0003415689adbb47326d
+```
+* VASPs store the mapping internally
+
+## User Story:
```
davidiw, line 146: prefer this to be really, really early like it was

also realized in reading and writing this, that we should be consistent in all our docs:
Bob's VASP is VASP B
Alice's VASP is VASP A

And we should probably have a pre-conditions to the user story:
* Alice's VASP, VASP A, has registered a DiemID domain at avasp
* Bob's VASP, VASP B, has registered a DiemID domain at bvasp

davidiw: but that gets deep into the weeds... hmm... something to chew on... technical versus overview

dips/dip-10.md
```diff
@@ -101,130 +92,62 @@ struct DiemIdDomainEvent {
...
-
-## Profile Picture URI
-Profile picture URI is a URI that represents a profile picture of a user. Wallet app must support two possible URI schemas for profile picture.
-
-* **Https schema.** If URI starts with https:, wallet must attempt to fetch profile picture from the HTTP URI provided. When fetching this URI wallet app will not provide any authentication. Wallet app will use GET method to fetch the profile picture specified using https schema.
-  * Ex.: *https://novi.com/pictures/alice.png*
-* **Data schema.** If URI starts with data:, wallet must interpret the URI as a [Data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) that contains profile picture encoded.
-  * Ex.: *data:image/gif;base64,AAAAAAAA*
-* **Unencrypted http** schema must not be used to share profile pictures.
+coinbase, 0xf72589b71ff4f8d139674a3f7369c69b
```
davidiw, line 141: let's not put well known entities here

dips/dip-10.md
davidiw: Yeah, definitely believe we need to introduce this earlier so we can gloss over the technical details to make the rest of the pieces fit together

* User exchanges identifiers
* VASP does on-chain look up
* VASP performs off-chain protocol
* VASP submits on-chain transaction

davidiw: which then maps to the deep technical pieces

dips/dip-10.md
```diff
@@ -13,47 +14,37 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-Diem ID is a string in format `<user-id>$<diem-id-domain>`.
+Example: `alice@avasp`
 
-Example: `andrey$novi`
-
-* **user-id**, identifies an account within a VASP:
+* `user_identifier` is a unique identifier meaningful only to a VASP that represents either the source or destination of a transaction. In practice, they are likely to be re-used and should be capable of both being the sender and recipient in a transaction. Their specific lifetime is undefined but should ideally be meaningful to the owner of the identifier. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]*$`
+  * Valid regular expression: ^[a-zA-Z0-9][a-zA-Z0-9-_\.]*$
```
davidiw, line 31: I regret removing the +, Adeniyi reminded me an important use case, where I can use +... to indicate a memo line not sure if we want that, but worth poking @aylinaykol

dips/dip-10.md
```diff
@@ -13,47 +14,37 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
 
-## Diem ID domains
-Each Diem ID domain is associated with a single VASP. This information is stored on the block chain in a `DiemIdDomains` resource published in a parent VASP's on-chain account.
+
+# Naming Service (Off-Chain Lookup Service)
+The Diem ID specifies that VASPs can receive payments at a vasp_domain_identifier. The intent is to eventually store these values on-chain. As an initial effort into this space, the Diem Association stores a file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
+```
+avasp, 0xf72589b71ff4f8d139674a3f7369c69b
+bvasp, 0xc5ab123458df0003415689adbb47326d
+```
```
davidiw, line 44: would recommend we put this at the end of the on-chain, mentioning that initial efforts will leverage a CSV file until the on-chain data is ready

dips/dip-10.md
```diff
@@ -101,130 +92,62 @@ struct DiemIdDomainEvent {
...
-  * *For example*: VASP can define a 'contact list' and only share information if the sender is in receiver's contacts
-* **Payment endpoint** to perform actual money transfer.
+# ReferenceID Exchange
+The PaymentMetadata transactions require a ReferenceID in order to submit a transaction to the chain. PaymentMetadata transactions reveal nothing distinctly about either the sender or recipient and do not create a linkability between the sender or receiver across payments. Hence, the ReferenceID must be established off-chain and this protocol defines one such way to do so leveraging the PaymentAddresses.
 
-## Info endpoint
+## ReferenceID Command
+A VASP intending to send a payment from one VASP to another when leveraging Diem IDs constructs a ReferenceIDCommand submits that to the recipient side. If the recipient knows the recipient PaymentAddress and can potentially accept payments, it returns a success. In the case it cannot, it returns an error indicating the reason.
```
davidiw, line 100: fix naming I see both PaymentAddress and Diem ID, plus is it PayAddress?

davidiw (commented):
A lot to chew on, I'm not convinced I have the best wording or structure in mind, so consider this a conversation. Also what's up with the conflicts on the tests? weird :(

dips/dip-10.md
sunmilee: Pay Address, i just pushed a PR to fix all of it

dips/dip-10.md
sunmilee: ok agree a user flow at the beginning might be a good overview

dips/dip-10.md
```diff
@@ -101,130 +91,62 @@ struct DiemIdDomainEvent {
...
-
-In order to support peer-2-peer payments, VASPs need to implement two Diem ID endpoints:
-
-* **Info endpoint** for querying optional information about the receiver. VASPs can provide this endpoint to improve UX and security of payment, but they are not obligated to - VASPs may choose not to share any information about the user via this endpoint, including whether the user exists or not.
-  * *For example*: VASP can define a 'contact list' and only share information if the sender is in receiver's contacts
-* **Payment endpoint** to perform actual money transfer.
 
-## Info endpoint
+# ReferenceID Exchange
+The PaymentMetadata transactions require a ReferenceID in order to submit a transaction to the chain. PaymentMetadata transactions reveal nothing distinctly about either the sender or recipient and do not create a linkability between the sender or receiver across payments. Hence, the ReferenceID must be established off-chain and this protocol defines one such way to do so leveraging the PaymentAddresses.
```
xli, line 96: "PaymentAddresses"?

dips/dip-10.md
```diff
@@ -101,130 +91,62 @@ struct DiemIdDomainEvent {
...
 
-**Info endpoint** is used by the **Sender** to request information about the **Receiver** before submitting an actual payment.
+## ReferenceID Command
+A VASP intending to send a payment from one VASP to another when leveraging Pay Addresss constructs a ReferenceIDCommand submits that to the recipient side. If the recipient knows the recipient PaymentAddress and can potentially accept payments, it returns a success. In the case it cannot, it returns an error indicating the reason.
```
xli, line 99: PaymentAddress -> PayAddress?

dips/dip-10.md
```diff
@@ -101,130 +91,62 @@ struct DiemIdDomainEvent {
...
-|receiver_id   |str          | Y            | Diem ID of the receiver |
-
-**Response:**
-```json
 {
-   "receiver": "<User Object>"
+   "_ObjectType": "CommandRequestObject",
+    "command_type": "ReferenceIDCommand",
+    "command": {
+	    "_ObjectType": "ReferenceIDRequest",
```
xli, line 109: Should be same with command_type ReferenceIDCommand

dips/dip-10.md
```diff
@@ -101,130 +91,62 @@ struct DiemIdDomainEvent {
...
-**Response:**
-```json
 {
-   "receiver_address": "<Receiver VASP on-chain account address>",
-   "reference_id": "<Payment_Reference_ID>",
-   "receiver": "<User Object>"
+   "_ObjectType": "CommandResponseObject",
+    "status": "success",
+    "result": {
+	    "_ObjectType": "ReferenceIDResponse",
```
xli, line 125: Suggest to use ReferenceIDCommandResponse

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-  * Diem ID provides better privacy compared to using sub-addresses, because Diem ID is not shared on the chain and will always use a single-use identifier for the on-chain ID
-  * From a user's perspective, Diem ID is a persistent identifier.  But from the perspective of the chain, Diem IDs do not exist and instead transactions are have a single-use ID to identify them. By contrast, from the viewpoint of both a user and the chain, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy, but appears on-chain and could thus be used to link payments if it isn't recycled upon each usage.
+Pay Address provides a convenient method for identifying users within a VASP.
+* For peer-2-peer payments, Pay Address allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Pay Address once. In this case Pay Address plays the role of an email address for payments. Benefits for this use case:
```
davidiw, line 18: No point for this bullet, would recommend we just extend the previous sentence.

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 
 # Summary
 
-This DIP describes Diem ID - the human-readable identifier for user accounts.   
-
-In addition to describing the format of the Diem ID, this DIP also describes the off-chain protocol for the peer-2-peer payment with Diem ID.
+This DIP describes Pay Address - the human-readable identifier for user accounts.
```
davidiw, line 14: I will not argue about this, but is this where we landed with @aylinaykol

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+bvasp, 0xc5ab123458df0003415689adbb47326d
+```
 
-# User story
-Below is an example of using Diem ID domain for transferring money from one user to another. 
+## User Story: 
+* Bob wants to receive funds from Alice
+* Bob registers a Pay Address at VASP B: `bob@bvasp`
+* Alice's Pay Address is registered at VASP A: `alice@avasp`
+* Bob shares either this Pay Address, `bob@bvasp`, or a PaymentURI, `diem://bob@bvasp` with Alice
```
davidiw, line 36: let's delete the PaymentURI for now.

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* Alice logs into VASP A, enters Bob’s identifier, an amount to pay, and submits payment
+* Alice’s VASP (VASP A) contacts Bob’s VASP’s (VASP B) off-chain API with the sender identifier `alice@avasp` and requests a reference_id, `rb1`
+* VASP A constructs a transaction with the specified amount and the reference_id `rb` and submits it to the diem network
```
davidiw, line 39: rb versus rb1

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* VASP B receives a transaction with metadata containing `rb1`, deposits the amount in his account, and attaches the relevant metadata (e.g., Alice’s Pay Address) to the transaction so Bob can confirm the transaction
```
davidiw, line 40: his? Bob's?

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-* Sender's wallet attempts to fetch the receiver's profile picture and full name along with a confirmation dialog
-* Sender confirms the identity of the recipient
-* Sender presses ‘send’ button
-* Receiver's account is credited the amount in the transaction
-* Receiver's wallet displays recent transactions and their senders
 
-# Diem ID format
-Diem ID is a string in format `<user-id>$<diem-id-domain>`.
+# Pay Address Format
+Diem defines globally unique identifiers to facilitate transacting on the Diem Payment Network. As Diem stores no PII on-chain, these identifiers are exchanged during the off-chain process between two VASPs to identify a distinct source and destinations within their respective VASPs in order to produce a reference_id that can be stored on-chain. The format for the identifiers follows:
```
davidiw, line 44: define PII ... might also be worth describing goals and properties before or after the user story

goals include:
* Eliminate the need for PII on-chain for peer-to-peer transactions
* Provide persistent, human-readable identifiers for sending and receiving transactions
any more?

dips/dip-10.md
```diff
@@ -11,49 +11,60 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+```
+avasp, 0xf72589b71ff4f8d139674a3f7369c69b
+bvasp, 0xc5ab123458df0003415689adbb47326d
+```
 
 # On-chain data
+## Pay Address Domain
 
 ```
 resource struct DiemIdDomains {
```
davidiw, line 70: should we call these PayAddressDomains :P

dips/dip-10.md
```diff
@@ -65,21 +76,21 @@ struct DiemIdDomain {
 }
 ```
 * Fields definition:
-   * `domain` - name of a Diem ID domain 
+   * `domain` - name of a Pay Address domain 
 * The `DiemIdDomains` resource can only be published into an account with `Role == PARENT_VASP_ROLE_ID`.
-* The `DiemIdDomains` contains a list of `DiemIdDomain` structs that are associated with a VASP. As such, it is possible to register more than one Diem ID Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
+* The `DiemIdDomains` contains a list of `DiemIdDomain` structs that are associated with a VASP. As such, it is possible to register more than one Pay Address Domain for a given VASP. We are providing this opportunity to support possible situations when two VASPs merge into one, or when a company wants to provide multiple wallet apps, while having only a single VASP parent account on the chain.
```
davidiw, line 81: I dunno if w need to say why more than one would be beneficial... just simply stating that a VASP can potentially have 1 or more PayAddressDomains

dips/dip-10.md
```diff
@@ -65,21 +76,21 @@ struct DiemIdDomain {
...
   * Only TC account can create and publish `DiemIdDomains` resource
   * Only TC account can add, remove or update an `DiemIdDomain` within `DiemIdDomains` resource
-* In order to register a Diem Id domain, a VASP needs to submit a request to Diem Association. LA will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem Id Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
-* The Diem ID domain can be created as part of creating a `ParentVASP` account.
-* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
+* In order to register a Pay Address domain, a VASP needs to submit a request to Diem Association. LA will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Pay Address Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
```
davidiw, line 85: LA or DA and define it -- Diem Association (DA).

dips/dip-10.md
```diff
@@ -65,21 +76,21 @@ struct DiemIdDomain {
...
+* The Pay Address domain can be created as part of creating a `ParentVASP` account.
```
davidiw, line 86: let's be clear here, it is expected to be created during the construction of a ParentVASP and existing ParentVASP accounts will have the resource instantiated by the DiemRoot account.

dips/dip-10.md
```diff
@@ -65,21 +76,21 @@ struct DiemIdDomain {
...
+* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Pay Addresss and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
```
davidiw, line 87: no, let's strike this,

dips/dip-10.md
```diff
@@ -65,21 +76,21 @@ struct DiemIdDomain {
...
 
-## Diem ID domain events
+## Pay Address Domain Events
```
davidiw, line 89: here's where we state that an application can be built to monitor for these events to construct a mapping of pay address domain to vasp account

dips/dip-10.md
```diff
@@ -101,130 +112,40 @@ struct DiemIdDomainEvent {
...
-In order to support peer-2-peer payments, VASPs need to implement two Diem ID endpoints:
 
-* **Info endpoint** for querying optional information about the receiver. VASPs can provide this endpoint to improve UX and security of payment, but they are not obligated to - VASPs may choose not to share any information about the user via this endpoint, including whether the user exists or not.
-  * *For example*: VASP can define a 'contact list' and only share information if the sender is in receiver's contacts
-* **Payment endpoint** to perform actual money transfer.
+# ReferenceID Exchange
+The PaymentMetadata transactions require a ReferenceID in order to submit a transaction to the chain. PaymentMetadata transactions reveal nothing distinctly about either the sender or recipient and do not create a linkability between the sender or receiver across payments. Hence, the ReferenceID must be established off-chain and this protocol defines one such way to do so leveraging Pay Address.
 
-## Info endpoint
+## ReferenceID Command
```
davidiw, line 119: no need for this extra header

davidiw (commented):
Okay, this is coming together really well, I think the next push should probably make this ready for more reviews

dips/dip-10.md
sunmilee: Are the goals listed in #Motivation section not enough? I can also just elaborate more there

davidiw: Hmm... this feels a little redundant, but whatever... :)

dips/dip-10.md
```diff
@@ -67,19 +77,18 @@ struct DiemIdDomain {
...
-* The Diem ID domain can be created as part of creating a `ParentVASP` account.
-* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
+* In order to register a Diem ID domain, a VASP needs to submit a request to Diem Association. Diem Association will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem ID Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
+* `DiemIDDomains` resource will be created as part of creating a `ParentVASP` account resource, and existing `ParentVASP` accounts without `DiemIDDomains` will have the resource instantiated by the DiemRoot account.
```
tzakian, line 85: A couple questions/comments here:
* I think it would be helpful to say what state the `DiemIDDomains` resource will be created in (i.e., empty). Or, does a VASP need to have at least one `DiemIDDomain` allocated to it at the time of creation?
* For existing vasps, they'll need to send a transaction to publish an empty `DiemIDDomains` resource before the association can allocate `DiemIDDomains` to it unless an administration script is used, however that seems pretty heavyweight. Relatedly, should this be the DiemRoot, or TreasuryCompliance account? It seems like it should be TreasuryCompliance to agree with the statement on line 80.

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+```
+avasp, 0xf72589b71ff4f8d139674a3f7369c69b
+bvasp, 0xc5ab123458df0003415689adbb47326d
+```
+
+## User Story: 
+* Bob wants to receive funds from Alice
+* Bob registers a Diem ID at VASP B: `bob@bvasp`
+* Alice's Diem ID is registered at VASP A: `alice@avasp`
+* Bob shares either this Diem ID, `bob@bvasp` with Alice
```
sblackshear, line 35: "either this Diem ID", or ...

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* Alice logs into VASP A, enters Bob’s identifier, an amount to pay, and submits payment
+* Alice’s VASP (VASP A) contacts Bob’s VASP’s (VASP B) off-chain API with the sender identifier `alice@avasp` and requests a reference_id, `rb1`
+* VASP A constructs a transaction with the specified amount and the reference_id `rb` and submits it to the diem network
```
sblackshear, line 38:
```suggestion
* VASP A constructs a transaction with the specified amount and the reference_id `rb` and submits it to the Diem network
```

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* VASP B receives a transaction with metadata containing `rb`, deposits the amount in Bob's account, and attaches the relevant metadata (e.g., Alice’s Diem ID) to the transaction so Bob can confirm the transaction
```
sblackshear, line 39: Perhaps specify where the attachment happens (inside the VASP's internal DB, presumably?)

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-Diem ID is a string in format `<user-id>$<diem-id-domain>`.
+# Diem ID Format
+Diem defines globally unique identifiers to facilitate transacting on the Diem Payment Network. As Diem stores no PII on-chain, these identifiers are exchanged during the off-chain process between two VASPs to identify a distinct source and destinations within their respective VASPs in order to produce a reference_id that can be stored on-chain. The format for the identifiers follows:
+`[user_identifier]@[vasp_domain_identifier]`
 
-Example: `andrey$novi`
+Example: `alice@avasp`
 
-* **user-id**, identifies an account within a VASP:
+* `user_identifier` is a unique identifier meaningful only to a VASP that represents either the source or destination of a transaction. In practice, they are likely to be re-used and should be capable of both being the sender and recipient in a transaction. Their specific lifetime is undefined but should ideally be meaningful to the owner of the identifier. Specification:
```
sblackshear, line 48: This description is somewhat hard to understand (e.g., both "unique" and "likely to be re-used"). Would something short like "`user_identifier` is a reusable identifier that represents either the source or destination end-user of a payment transaction" convey the same information (and if not, what's missing)?

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]+$`
+  * Valid regular expression: [a-zA-Z0-9]*
   * Maximum length: 63 characters
-  * Association between diem id and VASP is published on the chain 
 
-## Diem ID domains
-Each Diem ID domain is associated with a single VASP. This information is stored on the block chain in a `DiemIdDomains` resource published in a parent VASP's on-chain account.
+
+# Naming Service (Off-Chain Lookup Service)
+The Diem ID specifies that VASPs can receive payments at a vasp_domain_identifier. The intent is to eventually store these values on-chain. As an initial effort into this space, the Diem Association stores a file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
```
sblackshear, line 59: > As an initial effort into this space, the Diem Association stores a file mapping domain name to account address.

Can we specify where to find this mapping (or add a TODO if we don't know yet)?

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
-  * Diem ID provides better privacy compared to using sub-addresses, because Diem ID is not shared on the chain and will always use a single-use identifier for the on-chain ID
-  * From a user's perspective, Diem ID is a persistent identifier.  But from the perspective of the chain, Diem IDs do not exist and instead transactions are have a single-use ID to identify them. By contrast, from the viewpoint of both a user and the chain, a sub-address is an ephemeral user identifier that should be recycled frequently to preserve privacy, but appears on-chain and could thus be used to link payments if it isn't recycled upon each usage.
+Diem ID provides a convenient method for identifying users within a VASP. For peer-2-peer payments, Diem ID allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
```
sblackshear, line 17:
```suggestion
Diem ID provides a convenient method for identifying users within a VASP. Diem ID allows users to exchange human-readable identifiers as either the sender or the recipient of peer-to-peer payments. Instead of sending a [dip-5 address](https://github.com/diem/dip/blob/master/dips/dip-5.md) (that ideally needs to be refreshed for each payment), the receiver can just share their Diem ID once. In this case Diem ID plays the role of an email address for payments. Benefits for this use case:
```

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+* Privacy: There are concerns around leakage of personally identifiable information (PII) when using subaddresses. Diem ID provides better privacy compared to sub-addresses, because unlike sub-addresses, Diem ID is intended to be used in tandem with an off-chain protocol. Hence, Diem ID eliminates the need to expose PII on the chain 
+* Persistent Identifiers: Currently there is a lack of persistent user identifiers in the Diem ecosystem. while subaddresses need to be refreshed for each payment, Diem ID is a persistent identifier from a user's perspective. From the perspective of the chain, Diem IDs does not exist.
```
sblackshear, line 19:
```suggestion
* Persistent Identifiers: Currently there is a lack of persistent user identifiers in the Diem ecosystem. While subaddresses need to be refreshed for each payment, Diem ID is a persistent identifier from a user's perspective. From the perspective of the chain, Diem IDs does not exist.
```

dips/dip-10.md
sblackshear, line 17: I think the contrast with DIP5 here and in the bullets distracts from the explanation of what DiemID is (especially for readers not familiar with DIP5, which will hopefully be the majority in the future!). It might be clearer to explain what DiemID is in isolation, then add a later subsection on comparison with DIP5?

davidiw: I like your vision.

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+`[user_identifier]@[vasp_domain_identifier]`
 
-Example: `andrey$novi`
+Example: `alice@avasp`
 
-* **user-id**, identifies an account within a VASP:
+* `user_identifier` is a unique identifier meaningful only to a VASP that represents either the source or destination of a transaction. In practice, they are likely to be re-used and should be capable of both being the sender and recipient in a transaction. Their specific lifetime is undefined but should ideally be meaningful to the owner of the identifier. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]*$`
+  * Valid regular expression: [a-zA-Z0-9]*
```
xli, line 50: `*` means 0-any, it probably should be `+`, notice previous version has 1 char [a-zA-Z0-9] at beginning, hence `*` at the end is fine.  
why we change this regex? I feel original one is fine and better for including `-_+.`

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
   * Maximum length: 64 characters
-  * Stored internally in the VASP. Not published on the chain, there is no way for an external entity to list all the users within VASP.
-  * Defined by the VASP - user selected, auto-generated, etc.
-* **diem-id-domain** identifies a VASP:
+* `vasp_domain_identifier` is a well-known string that can be mapped to a VASP. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]+$`
+  * Valid regular expression: [a-zA-Z0-9]*
```
xli, line 54: same here, this regex means empty string is valid.

dips/dip-10.md
```diff
@@ -101,130 +110,39 @@ struct DiemIdDomainEvent {
...
-   "receiver": "<User Object>"
+   "_ObjectType": "CommandResponseObject",
+    "status": "success",
+    "result": {
+	    "_ObjectType": "ReferenceIDCommandResponse",
+	    "recipient_address": "c5ab123458df0003415689adbb47326d",
+    },
+    "cid": "12ce83f6-6d18-0d6e-08b6-c00fdbbf085a",
 }
 ```
```
xli, line 181: We don't have a table to describe each field anymore? I think it worth the effort to give simple description, instead of asking reader to derive them from the example code

dips/dip-10.md
sunmilee: it does yeah, i think it is useful to point out it should be unique to a user though?

dips/dip-10.md
sunmilee: Discussed with Aylin to only have alpha numeric code but we may add other symbols in later

dips/dip-10.md
sunmilee: This is for info and pay endpoint which are not in scope for this DIP

dips/dip-10.md
```diff
@@ -101,130 +110,55 @@ struct DiemIdDomainEvent {
...
-
-Additionally, all fields in this object are informational - **they must not be used for KYC and only intended for UI purposes**.
-```json
-{
-   "diem_id": "<Diem_ID>",
-   "display_name": "<Display name>",
-   "profile_picture": "<profile_picture_uri>"
+If the amount is below the travel rule limit, the sending VASP can send a p2p transaction with PaymentMetadata and the `reference_id` VASPs agreed on to settle the transaction. 
+```
+enum PaymentMetadata {
```
sunmilee, line 152: @davidiw should I put this here? Are we good to go with paymentmetadata metadata type?

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+# Diem ID Format
+Diem defines globally unique identifiers to facilitate transacting on the Diem Payment Network. As Diem stores no PII on-chain, these identifiers are exchanged during the off-chain process between two VASPs to identify a distinct source and destinations within their respective VASPs in order to produce a reference_id that can be stored on-chain. The format for the identifiers follows:
+`[user_identifier]@[vasp_domain_identifier]`
 
-Example: `andrey$novi`
+Example: `alice@avasp`
 
-* **user-id**, identifies an account within a VASP:
+* `user_identifier` is a reusable identifier that represents either the source or destination end-user of a payment transaction. It is unique to per user at VASP level. Specification:
   * Case insensitive
```
bmwill, line 49: If these identifiers are case insensitive then why don't we just enforce that they're lowercase?

dips/dip-10.md
```diff
@@ -1,7 +1,7 @@
 ---
 dip: 10
 title: Diem ID Spec
-authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb)
+authors: Andrey Chursin(@andll), Kevin Hurley (@kphfb), Sunmi Lee (@sunmilee), David Wolinsky (@davidiw)
```
davidiw, line 4: Please re-order:

SL, DW, AC, KH

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+bvasp, 0xc5ab123458df0003415689adbb47326d
+```
+
+## User Story: 
+* Bob wants to receive funds from Alice
+* Bob registers a Diem ID at VASP B: `bob@bvasp`
+* Alice's Diem ID is registered at VASP A: `alice@avasp`
+* Bob shares the Diem ID, `bob@bvasp` with Alice
+* Alice logs into VASP A, enters Bob’s identifier, an amount to pay, and submits payment
+* Alice’s VASP (VASP A) contacts Bob’s VASP’s (VASP B) off-chain API with the sender identifier `alice@avasp` and requests a reference_id, `rb1`
```
davidiw, line 37: rb or rb1?

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+
+# Naming Service (Off-Chain Lookup Service)
+The intent is to eventually store Diem ID domains on-chain; however, the Diem Association will initially store and distribute a CSV file mapping domain name to account address. The account address should be a parent VASP. The VASP sending a payment can then obtain the DualAttestation::Credentials::base_url to perform the ReferenceID exchange. An example CSV is:
+```
+avasp, 0xf72589b71ff4f8d139674a3f7369c69b
+bvasp, 0xc5ab123458df0003415689adbb47326d
+```
 
 # On-chain data
+## Diem ID Domain
```
bmwill, line 66: This part of the spec seems to omit defining an easy way to go from Domain -> Account Address. Looking at the "Events" section below also doesn't seem to address this fully since you'd have to look at the stream from the beginning of time in order to reconstruct the up-to-date mapping of Domain -> Account Address.

dips/dip-10.md
xli: I mean the attributes table for ReferenceIDCommand, we need a definition for it and looks like we are defining it in this DIP

xli: similarly, we also need an ReferenceIDCommandResponse attributes table.

dips/dip-10.md
```diff
@@ -101,130 +110,55 @@ struct DiemIdDomainEvent {
...
-   "reference_id": "<Payment_Reference_ID>",
-   "receiver": "<User Object>"
+   "_ObjectType": "CommandResponseObject",
+    "status": "success",
+    "result": {
+	    "_ObjectType": "ReferenceIDCommandResponse",
+	    "recipient_address": "c5ab123458df0003415689adbb47326d",
+    },
+    "cid": "12ce83f6-6d18-0d6e-08b6-c00fdbbf085a",
 }
```
xli, line 147: We need the attributes table for ReferenceIDCommand and ReferenceIDCommandResponse, unless we are going to create another DIP for defining them.

dips/dip-10.md
davidiw: Hmm..... it doesn't change the internal representation... my gut feeling is we're going to want more expressive character-set that maps to what at least the email standard is.

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+`[user_identifier]@[vasp_domain_identifier]`
 
-Example: `andrey$novi`
+Example: `alice@avasp`
 
-* **user-id**, identifies an account within a VASP:
+* `user_identifier` is a reusable identifier that represents either the source or destination end-user of a payment transaction. It is unique to per user at VASP level. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]*$`
+  * Valid regular expression: [a-zA-Z0-9]+
```
davidiw, line 50: add ` ` around the regex

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
   * Maximum length: 64 characters
-  * Stored internally in the VASP. Not published on the chain, there is no way for an external entity to list all the users within VASP.
-  * Defined by the VASP - user selected, auto-generated, etc.
-* **diem-id-domain** identifies a VASP:
+* `vasp_domain_identifier` is a unique string that is mapped to a VASP. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]+$`
+  * Valid regular expression: [a-zA-Z0-9]+
```
davidiw, line 54: add ` ` around the regex

dips/dip-10.md
```diff
@@ -101,130 +110,55 @@ struct DiemIdDomainEvent {
...
 ```
 
-| Field 	      | Type  | Required? 	| Description 	           |
-|-------	      |------ |-----------	|-------------	           |
-|diem_id         |str    | Y           | Diem ID of the user |
-|display_name     |str    | N           | Display name of the user. Usually a full name |
-|profile_picture  |str    | N           | Profile picture URL(see below) |
-
-If a user does not exist, VASP can choose to return an HTTP error, or the empty User Object, depending on it’s privacy choices.
+If the amount exceeds the travel rule limit, VASP must trigger an off-chain protocol for KYC exchange. The same `reference_id` must be used to perform a TR as described in [DIP-1](https://github.com/diem/dip/blob/master/dips/dip-1.md).
```
davidiw, line 158: YES, MUST

dips/dip-10.md
```diff
@@ -101,130 +110,55 @@ struct DiemIdDomainEvent {
...
-  * Ex.: *https://novi.com/pictures/alice.png*
-* **Data schema.** If URI starts with data:, wallet must interpret the URI as a [Data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) that contains profile picture encoded.
-  * Ex.: *data:image/gif;base64,AAAAAAAA*
-* **Unencrypted http** schema must not be used to share profile pictures.
+# Diem ID or Subaddress?
+[DIP-5](https://github.com/diem/dip/blob/master/dips/dip-5.md) defines subaddresses as a way of defining a user address. Between subaddress and Diem ID, the recommended way of representing a user address is to use a Diem ID, as Diem ID provides clear benefits over subaddresses:  
+* Subaddresses are ideally one-time use addresses, and need to be refreshed for each payment. However, a Diem ID can be shared once and will be a persistent identifier from a user's perspective.
+* There are concerns around leakage of PII when using subaddresses, because the subaddress has to be included in the on-chain transaction metadata. Diem ID provides better privacy, because Diem ID is exchanged off-chain.
```
davidiw, line 164: interesting... I'm not sure this is necessary, we're not writing an editorial, maybe strike and save for the blog update :P

davidiw (commented):
after tweaks, ready to approve

dips/dip-10.md
davidiw: yeah, we need the tables!!

dips/dip-10.md
sunmilee: not sure what you mean by the easy way to go from id to account address?

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+Diem defines globally unique identifiers to facilitate transacting on the Diem Payment Network. As Diem stores no PII on-chain, these identifiers are exchanged during the off-chain process between two VASPs to identify a distinct source and destinations within their respective VASPs in order to produce a reference_id that can be stored on-chain. The format for the identifiers follows:
+`[user_identifier]@[vasp_domain_identifier]`
 
-* **user-id**, identifies an account within a VASP:
+Example: `alice@avasp`
+
+* `user_identifier` is a reusable identifier that represents either the source or destination end-user of a payment transaction. It is unique to per user at VASP level. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-_+.]*$`
+  * Valid regular expression: `[a-zA-Z0-9]+`
```
CapCap, line 50: why did we remove the few special characters from this?

dips/dip-10.md
davidiw: I don't think there technically is an easy way, but there are two technical approaches:
* iterate across all accounts and an identify those that have this
* iterate across all events and identify updates or accounts along the way

dips/dip-10.md
```diff
@@ -67,19 +77,18 @@ struct DiemIdDomain {
...
 
-## Diem ID domain events
+## Diem ID Domain Events
 
 The Move contract that manages Diem ID domains must emit an event every time Diem ID domain is created, removed or updated. Those events are critical for applications to be able to efficiently index existing Diem ID domains.
-
-While Diem ID domains are published into VASP accounts, Diem ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow single event stream.
+An application can be built to listen for these events to construct a mapping of Diem ID domains to VASP accounts for lookup of onchain addresses given a Diem ID domain.
+While Diem ID domains are published into VASP account resource, Diem ID domain events are published under the Treasury Compliance account. We consolidate events under single account to allow indexers to follow a single event stream.
```
davidiw, line 91: @bmwill

davidiw (approved):
Looks good to me!

davidiw:
Let's make an issue for this DIP where folks can continue to give feedback.

dips/dip-10.md
bmwill: This doesn't quite solve my concerns with this approach. I've worried, that given how large our blockchain is that it'll be impossible to construct the canonical list without needing to walk every account or walk the whole history

davidiw: There are a set of events emitted by the TC, so you just need to get the unique accounts from that and query their current state or alternatively mutate them as you read the events.

dahliamalkhi (commented):
Thanks for making substantial improvements to this DIP!

I wanted to offer some high level comments:

* I would add a brief introduction to the effect of: 

  This DIP has two components. 

  The first part is a 2-level namespace for VASP user addresses, consisting of a username and a VASP name. It includes a method for registering  VASPs name(s) on the Diem blockchain. 

  The second part is a standard framework for initiating a payment transfer to a recipient whose address has already been communicated to an originator (by means left outside the scope of this DIP). The protocol standard arranges for the originator and recipient VASPs to coordinate a transaction referenceID. Other off-chain information exchanges can occur as described in other DIPs (e.g., in DIP-1) and can be embedded within the proposed framework. 

* I would refrain from claims about greater privacy, and any other unsubstantiated privacy statements  

* I am not supportive of the (two part) name Diem ID, since it it not clear it will be the only identity we will have on Diem. I would prefer something more specific, e.g., VAEP (VASP Endpoint) or the like.

* The user story is cute, but we should emphasize this is only for demonstration purposes, it is NOT part of the proposed standard

* I wonder whether we should permit more characters in VAEP in order to be compatible with addresses on other blockchains?

davidiw:
@dahliamalkhi, great feedback, routing it to: https://github.com/diem/dip/issues/156

dips/dip-10.md
```diff
@@ -67,19 +77,18 @@ struct DiemIdDomain {
...
   * Only TC account can create and publish `DiemIdDomains` resource
   * Only TC account can add, remove or update an `DiemIdDomain` within `DiemIdDomains` resource
-* In order to register a Diem Id domain, a VASP needs to submit a request to Diem Association. LA will perform certain checks (out of scope of this document) before issuing an on-chain transaction to register a Diem Id Domain. These checks intend to mitigate irrelevant claims and enforce uniqueness
-* The Diem ID domain can be created as part of creating a `ParentVASP` account.
-* An entity (potentially the Association) may at some point choose to expose an open source application(*Indexer*) that will allow indexing available Diem IDs and could provide a convenient REST API for applications to fetch information about the domains. The API of such an indexer is out of scope of this RFC.
+* `DiemIDDomains` resource will be created in an empty state as part of creating a `ParentVASP` account resource, and existing `ParentVASP` accounts without `DiemIDDomains` will have the resource instantiated by the DiemRoot account.
```
LuZhang-Lou, line 84: noob question: why not add the resources for existing ParentVASPs?

dips/dip-10.md
LuZhang-Lou: agree with @davidiw , one should be able to get a snapshot (by looking at the accounts that one feels interested in), and then traverse the later events from that point, to keep up with the latest state

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
-* **Payment endpoint** to perform actual money transfer.
+# ReferenceID Exchange
+The PaymentMetadata transactions require a ReferenceID in order to submit a transaction to the chain. PaymentMetadata transactions reveal nothing distinctly about either the sender or recipient and do not create a linkability between the sender or receiver across payments. Hence, the ReferenceID must be established off-chain and this protocol defines one such way to do so leveraging Diem ID.
 
-## Info endpoint
+A VASP intending to send a payment from one VASP to another when leveraging Diem IDs constructs a ReferenceIDCommand submits that to the recipient side. If the recipient knows the recipient Diem ID and can potentially accept payments, it returns a success. In the case it cannot, it returns an error indicating the reason.
 
-**Info endpoint** is used by the **Sender** to request information about the **Receiver** before submitting an actual payment.
+In the case that the amount to be sent would exceed the limit of the travel rule, the sending party should follow this exchange with a PaymentCommand using the same reference_id and specify the sending and receiving subaddresses as all 0.
```
LuZhang-Lou, line 119: we don't say anything about `PaymentCommand` in this dip? Is it the replacement for dip-1?

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
+**CommandObject:**
 
-**TODO: This endpoint is going to be merged into DIP 1**
+| Field 	    | Type 	     | Required? 	| Description 	           |
+|-------	    |------	     |-----------	|-------------	           |
+| _ObjectType   | str    | Y | Fixed value: `ReferenceIDCommand`|
+| sender        | str          | Y            | Sender's full Diem ID |
+| sender_address| str          | Y            | Sender's on-chain address |
+| recipient     | str          | Y            | Receiver's full Diem ID |
+| reference_id  | str          | Y            | Reference ID of this transaction to be included into payment metadata |
```
LuZhang-Lou, line 155: should we state that this reference id is also uuid formated? do we say this somewhere in this doc?

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
-```json
 {
-   "receiver_address": "<Receiver VASP on-chain account address>",
-   "reference_id": "<Payment_Reference_ID>",
-   "receiver": "<User Object>"
+   "_ObjectType": "CommandResponseObject",
+    "status": "success",
+    "result": {
+	    "_ObjectType": "ReferenceIDCommandResponse",
+	    "recipient_address": "c5ab123458df0003415689adbb47326d",
```
LuZhang-Lou, line 165: should we call this `receiver_address` to echo dip-1? either ways, we should be consistent with `recipient` v.s. `receiver`

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
-* Once the sender learns Payment_Reference_ID, it will submit an on-chain transaction including the `Payment_Reference_ID` as a metadata. Receiver will use the `Payment_Reference_ID` in the transaction to link the on-chain transaction to the payment.
+**CommandResponseObject:**
 
-* Sender will submit a transaction to the `receiver_address` on-chain account in the response. This account address can be either Parent VASP, or any of Child VASP accounts of the receiver VASP. This is needed to allow the receiver VASP to manage load between different on-chain accounts.
+| Field 	    | Type 	     | Required? 	| Description 	           |
+|-------	    |------	     |-----------	|-------------	           |
+| _ObjectType   | str        | Y | Fixed value: `CommandResponseObject`|
+| status       | str | Y | Either `success` or `failure`. |
+| result       | Result object | Y | The Result obejct of response. |
+| cid           | str         | Y            | A unique identifier for the Command. Should be a UUID according to [RFC4122](https://tools.ietf.org/html/rfc4122) with "-"'s included. |
```
LuZhang-Lou, line 178: why do we need `cid` in response? the sender should have request's cid when they get the response

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
 
-* `Payment_Reference_ID` must be unique for each payment, it must not be reused in any way to avoid leaking privacy information on-chain.
+**CommandResultObject:**
 
-* `Payment_Reference_ID` returned by this endpoint is not signed for KYC purposes. This means that payment above KYC threshold can not be submitted right away using the provided Payment_Reference_ID. Instead, sender VASP needs to proceed to KYC endpoint to negotiate the signed Reference ID.
+| Field 	    | Type 	     | Required? 	| Description 	           |
+|-------	    |------	     |-----------	|-------------	           |
+| _ObjectType   | str        | Y | Fixed value: `ReferenceIDCommandResponse`|
+| recipient_address       | str | Y | Receiver's on-chain address |
```
LuZhang-Lou, line 185: see comment above for `recipient` v.s. `receiver`

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
-{
-   "diem_id": "<Diem_ID>",
-   "display_name": "<Display name>",
-   "profile_picture": "<profile_picture_uri>"
+If the amount is below the travel rule limit, the sending VASP can send a p2p transaction with PaymentMetadata and the `reference_id` VASPs agreed on to settle the transaction. 
+```
+enum PaymentMetadata {
+    PaymentMetadataV0(ReferenceId),
 }
+type ReferenceId = [u8, 16];
```
LuZhang-Lou, line 192: do we want to remove this line?

dips/dip-10.md
```diff
@@ -101,130 +110,87 @@ struct DiemIdDomainEvent {
...
-|-------	      |------ |-----------	|-------------	           |
-|diem_id         |str    | Y           | Diem ID of the user |
-|display_name     |str    | N           | Display name of the user. Usually a full name |
-|profile_picture  |str    | N           | Profile picture URL(see below) |
-
-If a user does not exist, VASP can choose to return an HTTP error, or the empty User Object, depending on it’s privacy choices.
-
-## Profile Picture URI
-Profile picture URI is a URI that represents a profile picture of a user. Wallet app must support two possible URI schemas for profile picture.
+If the amount exceeds the travel rule limit, VASPs should trigger an off-chain protocol for KYC exchange. The same `reference_id` must be used to perform a TR as described in [DIP-1](https://github.com/diem/dip/blob/master/dips/dip-1.md).
```
LuZhang-Lou, line 195: current dip-1's schema uses subaddress hence there is a gap. I suggest at this point we do not talk about Travel Rule in dip-10. We can have an extension dip for it when we are sure what we want for Travel Rule

LuZhang-Lou (commented):
I think we should also talk about what happens when a reference id is used multiples times. Do we expect the receiver to refund?

dips/dip-10.md
```diff
@@ -13,47 +13,57 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
...
+  * Valid regular expression: `[a-zA-Z0-9]+`
   * Maximum length: 64 characters
-  * Stored internally in the VASP. Not published on the chain, there is no way for an external entity to list all the users within VASP.
-  * Defined by the VASP - user selected, auto-generated, etc.
-* **diem-id-domain** identifies a VASP:
+* `vasp_domain_identifier` is a unique string that is mapped to a VASP. Specification:
   * Case insensitive
-  * Regex: `^[a-zA-Z0-9][a-zA-Z0-9-]+$`
+  * Valid regular expression: `[a-zA-Z0-9]+`
   * Maximum length: 63 characters
```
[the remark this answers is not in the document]
bmwill: 64 + @ + 63 = 128?

dips/dip-10.md
bmwill: No this defines the onchain structure of the reference id, which is a uuid in a compact form.

dips/dip-10.md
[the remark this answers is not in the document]
sunmilee: 64 including "@" i'll add in to elaborate

dips/dip-10.md
sunmilee: not sure what you mean, we will be adding resources for the existing vasps too

dips/dip-10.md
sunmilee: just echoing dip1

dips/dip-10.md
sunmilee: why?

PR #153, opened 2021-03-25
DIP16 Draft: Transaction Scripts

sblackshear:


dips/dip-16.md
```diff
@@ -0,0 +1,112 @@
...
+pub struct Script {
+    code: Vec<u8>,
+    ty_args: Vec<TypeTag>,
+    args: Vec<TransactionArgument>,
+}
+```
+
+Here, `code` is a Move bytecode program. The `ty_args` are the same as for script functions. The `args` are a list of `TransactionArgument`s instead of a list of BCS-encoded values.
+
+Only a restricted set of programs can be included in `code`. The allowed programs are specified in an on-chain allowlist published at at `0x1::DiemTransactionPublishingOption::DiemTransactionPublishingOption.script_allowlist`. The allowlist contains the sha3-256 hash of each program rather than the code itself.
```
bob-wilson, line 83: I just noticed the repeated "at at" here.

More importantly, should we mention that DiemRoot is exempted from the allowlist check?

sblackshear: Very good point!

dips/dip-16.md
```diff
@@ -0,0 +1,120 @@
...
+    /// A transaction that executes an existing script function published on-chain.
+    ScriptFunction(ScriptFunction),
+}
+```
+
+The two `TransactionPayload` variants relevant to this DIP are `ScriptFunction` (which contains a pointer to a function already published in an on-chain Move module) and `Script` (a standalone, single-function bytecode program).
+
+## Script signature requirements
+
+Both script functions and the single function in a transaction script bytecode file have the following restrictions. For a function` f<ability_params>(param_types): ret_types` :
```
davidiw, line 33:
```suggestion
Both script functions and the single function in a transaction script bytecode file have the following requirements. For a function` f<ability_params>(param_types): ret_types`:
```

dips/dip-16.md
```diff
@@ -0,0 +1,120 @@
...
+
+* The `ret_types` list is empty (i.e., the function does not return a value)
+* The `param_types` list begins with one or more `signer` types
```
davidiw, line 36: can you point to signer types?

dips/dip-16.md
```diff
@@ -0,0 +1,120 @@
...
+
+In the Move language, a function can be declared with four different visibility levels: `public`, `private`, `public(friend)`, and `public(script)`. *Any* function with `public(script)` visibility can be invoked by a transaction if it satisfies the script signature requirements explained below.
+
+A client can define a `ScriptFunction` transaction payload using the following struct:
+
+```
+pub struct ScriptFunction {
+    module: ModuleId,
+    function: Identifier,
+    ty_args: Vec<TypeTag>,
```
davidiw, line 56: are these defined anywhere? can we point to them?

dips/dip-16.md
```diff
@@ -0,0 +1,120 @@
...
+
+A transaction script is a Move bytecode program that contains a single function `f`. The signature of `f` must satisfy the script signature requirements described below.
+
+A client can define a `Script` transaction payload using the following struct.
+
+```
+pub struct Script {
+    code: Vec<u8>,
+    ty_args: Vec<TypeTag>,
+    args: Vec<TransactionArgument>,
```
davidiw, line 77: can you point to TransactionArgument?

davidiw (commented):
overall seems pretty reasonable, I would love to see pointers to other docs, so that this DIP could be included in some developer guide or specification on an end-to-end process.

Not sure what the point of the changelog is?

dips/dip-16.md
sblackshear: They are all defined in move-core-types. I'll add a note below.

sblackshear:
>Not sure what the point of the changelog is?

This is (I think, but am not sure) how we agreed to document which parts of a DIP apply to which versions of Diem. Here, it's a bit awkward because the DIP didn't exist before Diem v1, but it is describing both v1 and v2 features. Definitely open to a different way of conveying this information if we can all standardize on it.

davidiw (commented):
add links and let's land this?

dips/dip-16.md
```diff
@@ -0,0 +1,120 @@
...
+    /// A transaction that executes an existing script function published on-chain.
+    ScriptFunction(ScriptFunction),
+}
+```
+
+The two `TransactionPayload` variants relevant to this DIP are `ScriptFunction` (which contains a pointer to a function already published in an on-chain Move module) and `Script` (a standalone, single-function bytecode program).
+
+## Script signature requirements
+
+Both script functions and the single function in a transaction script bytecode file have the following requirements. For a function` f<ability_params>(param_types): ret_types`:
```
xli, line 33: I'm confused by this ability_params, because in our rust code, the ability is refer to copy, drop, store and key, is this ability_params same thing or they are generic type info?

PR #159, opened 2021-04-06
Add draft for DIP 169: Multi-agent Transactions

emmazzz:
This PR adds draft for DIP 169: muli-agent transactions.
Issue for discussion is here: #169

### DIFF
--- a/dips/dip-14.md
+++ b/dips/dip-14.md
@@ -1,7 +1,7 @@
 ---
 dip: 14
 title: Multi-Agent Transactions
-author: Emma Zhong(@emmazzz), Tim Zakian, Sam Blackshear
+authors: Emma Zhong(@emmazzz), Tim Zakian, Sam Blackshear
 Status: Draft
 type: Standard
 created: 04/05/2021
--- a/dips/dip-14.md
+++ b/dips/dip-14.md
@@ -1,0 +1,245 @@
+---
+dip: 14
+title: Multi-Agent Transactions
+author: Emma Zhong(@emmazzz), Tim Zakian, Sam Blackshear
+Status: Draft
+type: Standard
+created: 04/05/2021
+---
+
+
+# Summary
+
+This DIP describes a new scheme of user transaction: multi-agent transactions.
+
+# Abstract
+
+Currently in the Diem Framework, a single transaction acts on behalf of a single on-chain account. However, there is no mechanism for multiple on-chain accounts to agree on a single atomic transaction. This DIP presents a new scheme of transactions--multi-agent transactions--which act on behalf of multiple on-chain accounts. Multi-agent transactions leverage Move’s [_`signer`_](https://developers.diem.com/docs/move/move-signer/) type to allow essentially any arbitrary atomic actions in one transaction involving multiple on-chain accounts.
+
+## Terminology:
+
+* Primary signer: This is the account that the transaction is sent from. This account’s sequence number is incremented, and is the account that pays gas. There is precisely one of these for every transaction.
+* Secondary signer: This is any account that participates in a multi-signer/agent transaction that isn’t the primary sender. There can be 0-N of these. The 0 case is a normal transaction today. 
+
+# Motivation/Use Cases
+
+## Minting Directly to VASPs
+
+Today in the Diem Framework, we need two transactions in order to mint money through designated dealers to VASPs. The first transaction is a tiered-mint transaction sent by the treasury compliance account to mint money to a designated dealer’s account. And the second transaction transfers the minted funds from the designed dealer’s account into the VASP’s account. This procedure requires two transactions due to the restriction that each transaction can only have one signer argument. In the multi-agent scheme, this restriction no longer exists and we can perform both steps in a single atomic transaction(see code sample below). This atomicity implies that designated dealers no longer need custody services for the money they temporarily hold in their accounts, making it more affordable and much quicker to operate as a designated dealer. This should make it much easier for more DDs to onboard in the Diem ecosystem, providing more options for VASPs. It is highly likely that in the future, DDs primarily mint directly on requests from VASPs.
+
+```
+fun mint<Coin>(
+   tc: signer,
+   dd: signer,
+   vasp_addr: address,
+   amount: u64,
+   tier_index: u64
+) {
+   // First, TC mints to DD.
+   tiered_mint<Coin>(
+       tc, address_of(dd), amount, tier_index
+   );
+
+   // Then, DD distributes funds to VASP.
+   pay<Coin>(
+       dd, vasp_addr, amount
+   );
+}
+```
+
+
+
+## Atomic Swaps
+
+In order to do a currency exchange in the current scheme between two on-chain entities Alice and Bob, we need two transactions and possibly an escrow. The time gap between the two transactions can lead to potential problems such as resource lockup. With the new multi-agent transaction scheme, we only need one atomic transaction signed by both Alice and Bob, in which payments are sent in both directions(see the code sample below). In this case, both Alice and Bob have the same control over the transaction contents including exchange rates and expiration time.
+
+
+```
+// alice and bob agree on the values of amount_usd and amount_eur 
+// off-chain.
+fun exchange(
+    alice: signer, bob: signer, 
+    amount_usd: u64, amount_eur: u64
+) {
+    // First, Alice pays Bob in currency USD.
+    pay<USD>(alice, address_of(bob), amount_usd);
+
+    // Then, Bob pays Alice in currency EUR.
+    // Previously, this line has to be a separate transaction.     
+    pay<EUR>(bob, address_of(alice), amount_eur);
+}
+```
+
+
+
+## Actual Dual-attested Transactions 
+
+Currently, we execute a [_dual-attested payment_](https://dip.diem.com/dip-1/) transaction by including a signature of the payee as an argument to the transaction, and checking the validity of the signature on-chain if the payment is between two different VASPs and the amount is over a certain threshold. In order to accomplish that, we have to run some fairly complex code to reconstruct and verify the signature ad-hoc on-chain. With the new scheme, we can simply require that the payee signs the entire transaction as well, and verify the signatures of both the payer and payee while validating the transaction. The code samples below show the before and after of what such a transaction might look like. 
+
+Some benefits brought by the multi-agent scheme in this scenario are:
+
+* Extensibility: In the current scheme, our dual attestation protocol is very specific to travel rule payments. And extending this protocol to dual attestation on a different type of action would require another ad-hoc data format and signature scheme. The multi-agent transaction scheme generalizes this by allowing dual attestation on any action encodable by a Move script without having to define new data formats or signature schemes.
+* Security: Multi-agent scheme closes two security loopholes of our current protocol:
+    * Replayability: Since the payee only signs over the payer and the amount of the payment, today the payer can replay the transaction using old payee approvals. In the new scheme, this wouldn’t be possible since the payee has to sign over the sequence number of the transaction as well, disallowing any replay of the transaction.
+    * Currency: Today’s protocol does not let the payee commit to the currency type so the payer can reuse a payee approval for payment in one currency for payment in another with the same amount. Again, since the payee has to sign over the entire transaction including the currency type, this scenario would be impossible with the multi-agent scheme. 
+* Earlier failure: Currently a bad signature in a travel rule transaction will be caught at execution time, and the transaction will abort. In the new scheme, the signature is checked at validation time and the transaction will be discarded instead.
+
+```
+/// Before:
+fun dual_attested_payment<Coin>(
+    payer: signer, payee: address, amount: u64,
+    metadata: vector<u8>, 
+    metadata_signature: vector<u8>
+){
+    let msg = message(payer, amount, metadata);
+    verify_signature(compliance_key(payee), msg);
+    pay<Coin>(payer, payee, amount, metadata);
+}
+
+/// After:
+fun dual_attested_payment<Coin>(
+    payer: signer, payee: signer, amount: u64,
+    metadata: vector<u8>
+){
+    pay<Coin>(payer, address_of(payee), amount, metadata);
+}
+```
+
+
+
+## Arbitrary Atomic Actions in One Transaction
+
+Besides the use cases mentioned above, our new multi-agent transaction scheme allows the sender to encode any arbitrary combination of actions in a single atomic transaction. This added expressiveness opens up a lot more possibilities in the type of scenarios that we can implement. Some examples include [_delivery versus payment_](https://www.investopedia.com/terms/d/dvp.asp) (a crucial feature for securities settlement) and atomic administrative approval of sensitive actions (e.g., account creation). 
+
+
+
+
+# Code Changes to Support Multi-agent
+
+## DiemVM Code Changes
+
+### Add New Enum `RawTransactionWithData`
+
+In order to preserve backward compatibility, we can’t add more fields representing secondary signers to `RawTransaction` struct. Thus we add the following additional enum `RawTransactionWithData`, which the sender and secondary signers will sign over to authenticate the transaction.
+
+
+```
+pub enum RawTransactionWithData {
+    MultiAgent {
+        raw_txn: RawTransaction,
+        secondary_signer_addresses: Vec<AccountAddress>,
+    },
+}
+```
+
+
+
+### Add New Enum `AccountAuthenticator`
+
+Previously each transaction can only have one signer, which is the sender of the transaction. In the multi-agent scheme, with the ability to have multiple signers, the transaction can potentially have multiple authenticators from different accounts and of different schemes. An `AccountAuthenticator` serves as the authenticator for one account. And a `TransactionAuthenticator` can contain multiple `AccountAuthenticator`s, as shown in the next subsection.
+
+
+```
+pub enum AccountAuthenticator {
+    /// Single signature
+    Ed25519 {
+        public_key: Ed25519PublicKey,
+        signature: Ed25519Signature,
+    },
+    /// K-of-N multisignature
+    MultiEd25519 {
+        public_key: MultiEd25519PublicKey,
+        signature: MultiEd25519Signature,
+    },
+}
+```
+
+
+
+### Add New Variants to `TransactionAuthenticator`
+
+We have added two new variants to `TransactionAuthenticator` to differentiate single-agent and multi-agent transaction scheme. In the single-agent scheme, sender is the only signer and sender’s signature is verified against the `RawTransaction`. While in the multi-agent scheme, the sender and all the secondary signers have to sign over `RawTransactionWithData::MultiAgent{ raw_txn, secondary_signer_addresses }`, to make sure that all parties agree to transact with each other. 
+
+
+```
+pub enum TransactionAuthenticator {
+    /// Single signature
+    /// This variant is kept for backwards compatibility.
+    Ed25519 {
+        public_key: Ed25519PublicKey,
+        signature: Ed25519Signature,
+    },
+    /// K-of-N multisignature
+    /// This variant is kept for backwards compatibility.
+    MultiEd25519 {
+        public_key: MultiEd25519PublicKey,
+        signature: MultiEd25519Signature,
+    },
+    /// Single-agent transaction.
+    SingleAgent { sender: AccountAuthenticator },
+    /// Multi-agent transaction.
+    MultiAgent {
+        sender: AccountAuthenticator,
+        secondary_signer_addresses: Vec<AccountAddress>,
+        secondary_signers: Vec<AccountAuthenticator>,
+    },
+    ..
+}
+```
+
+
+
+### Additional Checks During Transaction Validation
+
+In addition to signature checking, we also perform the following additional checks on signed multi-agent transactions during transaction validation:
+
+* The number of secondary authenticators in the signed transaction is the same as the number of secondary signer addresses that all parties have signed over.
+* There are no duplicates in the signers. In other words, sender and all the secondary signers have distinct account addresses.
+
+
+
+
+## Diem Framework Changes
+
+### Account Existence Checking in Prologue
+
+Today we check in the prologue that the sender of the transaction has a `DiemAccount` resource under their address. With the ability to have multiple signers, we need to check that all of them have `DiemAccount` resources during the prologue.
+
+```
+let i = 0;
+while (i < num_secondary_signers) {
+    let secondary_address = *Vector::borrow(&secondary_signer_addresses, i);
+    assert(
+        exists_at(secondary_address), 
+        Errors::invalid_argument(PROLOGUE_EACCOUNT_DNE)
+    );
+    i = i + 1;
+};
+```
+
+
+
+### Authentication Key Checking in Prologue
+
+Currently in the prologue, we check that the hash of the sender’s public key is equal to the authentication key that the sender has on chain. In the new scheme, we will need to perform the same check for public keys of all the secondary signers as well. Therefore, two new arguments are added to the `script_prologue` function. They are `secondary_signer_addresses: vector<address> `and `secondary_signer_public_key_hashes: vector<vector<u8>>`. In the prologue, we add a loop going through the two vectors to perform the aforementioned check. 
+
+
+```
+let i = 0;
+while (i < num_secondary_signers) {
+    let secondary_address = *Vector::borrow(&secondary_signer_addresses, i);
+    let signer_account = borrow_global<DiemAccount>(secondary_address);
+    let signer_public_key_hash = *Vector::borrow(&secondary_signer_public_key_hashes, i);
+    assert(
+        signer_public_key_hash == *&signer_account.authentication_key,
+        Errors::invalid_argument(PROLOGUE_EINVALID_ACCOUNT_AUTH_KEY),
+    );
+    i = i + 1;
+};
+```
+
+
+
+
+
+
--- a/dips/dip-14.md
+++ b/dips/dip-14.md
@@ -1,8 +1,8 @@
 ---
 dip: 14
 title: Multi-Agent Transactions
-authors: Emma Zhong(@emmazzz), Tim Zakian, Sam Blackshear
-Status: Draft
+authors: Emma Zhong, Tim Zakian, Sam Blackshear
+status: Draft
 type: Standard
 created: 04/05/2021
 ---
@@ -138,6 +138,7 @@ pub enum RawTransactionWithData {
 
 Previously each transaction can only have one signer, which is the sender of the transaction. In the multi-agent scheme, with the ability to have multiple signers, the transaction can potentially have multiple authenticators from different accounts and of different schemes. An `AccountAuthenticator` serves as the authenticator for one account. And a `TransactionAuthenticator` can contain multiple `AccountAuthenticator`s, as shown in the next subsection.
 
+Notice that `AccountAuthenticator` has a multi-signature variant `MultiEd25519`. `MultiEd25519` is a signature scheme an account uses to generate signatures. This is different from multi-agent, which is a new type of transaction.
 
 ```
 pub enum AccountAuthenticator {
@@ -159,6 +160,8 @@ pub enum AccountAuthenticator {
 ### Add New Variants to `TransactionAuthenticator`
 
 We have added two new variants to `TransactionAuthenticator` to differentiate single-agent and multi-agent transaction scheme. In the single-agent scheme, sender is the only signer and sender’s signature is verified against the `RawTransaction`. While in the multi-agent scheme, the sender and all the secondary signers have to sign over `RawTransactionWithData::MultiAgent{ raw_txn, secondary_signer_addresses }`, to make sure that all parties agree to transact with each other. 
+
+`Ed25519` and `MultiEd25519` are signature schemes of a single account so we have moved them to a new enum `AccountAuthenticator`(see the previous section). They are kept in `TransactionAuthenticator` for backwards compatibility.
 
 
 ```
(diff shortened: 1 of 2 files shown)


dips/dip-14.md
```diff
@@ -0,0 +1,245 @@
...
+This DIP describes a new scheme of user transaction: multi-agent transactions.
+
+# Abstract
+
+Currently in the Diem Framework, a single transaction acts on behalf of a single on-chain account. However, there is no mechanism for multiple on-chain accounts to agree on a single atomic transaction. This DIP presents a new scheme of transactions--multi-agent transactions--which act on behalf of multiple on-chain accounts. Multi-agent transactions leverage Move’s [_`signer`_](https://developers.diem.com/docs/move/move-signer/) type to allow essentially any arbitrary atomic actions in one transaction involving multiple on-chain accounts.
+
+## Terminology:
+
+* Primary signer: This is the account that the transaction is sent from. This account’s sequence number is incremented, and is the account that pays gas. There is precisely one of these for every transaction.
+* Secondary signer: This is any account that participates in a multi-signer/agent transaction that isn’t the primary sender. There can be 0-N of these. The 0 case is a normal transaction today.
```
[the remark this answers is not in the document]
bob-wilson: I'm suggesting to avoid the "multi-signer" term here. This doc has consistently used the "multi-agent" term for the topic of this DIP, and there has often been confusion vs. the related idea of a "multi-sig" account.

dips/dip-14.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Motivation/Use Cases
+
+## Minting Directly to VASPs
+
+Today in the Diem Framework, we need two transactions in order to mint money through designated dealers to VASPs. The first transaction is a tiered-mint transaction sent by the treasury compliance account to mint money to a designated dealer’s account. And the second transaction transfers the minted funds from the designed dealer’s account into the VASP’s account. This procedure requires two transactions due to the restriction that each transaction can only have one signer argument. In the multi-agent scheme, this restriction no longer exists and we can perform both steps in a single atomic transaction(see code sample below). This atomicity implies that designated dealers no longer need custody services for the money they temporarily hold in their accounts, making it more affordable and much quicker to operate as a designated dealer. This should make it much easier for more DDs to onboard in the Diem ecosystem, providing more options for VASPs. It is highly likely that in the future, DDs primarily mint directly on requests from VASPs.
```
bob-wilson, line 28:
```suggestion
Today in the Diem Framework, we need two transactions in order to mint money through designated dealers to VASPs. The first transaction is a tiered-mint transaction sent by the treasury compliance account to mint money to a designated dealer’s account. The second transaction transfers the minted funds from the designed dealer’s account into the VASP’s account. This procedure requires two transactions due to the restriction that each transaction can only have one signer argument. In the multi-agent scheme, this restriction no longer exists and we can perform both steps in a single atomic transaction (see code sample below). This atomicity implies that designated dealers no longer need custody services for the money they temporarily hold in their accounts, making it more affordable and much quicker to operate as a designated dealer. This should make it much easier for more DDs to onboard in the Diem ecosystem, providing more options for VASPs. It is highly likely that in the future, DDs primarily mint directly on requests from VASPs.
```

dips/dip-14.md
```diff
@@ -0,0 +1,245 @@
...
+       dd, vasp_addr, amount
+   );
+}
+```
+
+
+
+## Atomic Swaps
+
+In order to do a currency exchange in the current scheme between two on-chain entities Alice and Bob, we need two transactions and possibly an escrow. The time gap between the two transactions can lead to potential problems such as resource lockup. With the new multi-agent transaction scheme, we only need one atomic transaction signed by both Alice and Bob, in which payments are sent in both directions(see the code sample below). In this case, both Alice and Bob have the same control over the transaction contents including exchange rates and expiration time.
```
bob-wilson, line 54:
```suggestion
In order to do a currency exchange in the current scheme between two on-chain entities Alice and Bob, we need two transactions and possibly an escrow. The time gap between the two transactions can lead to potential problems such as resource lockup. With the new multi-agent transaction scheme, we only need one atomic transaction signed by both Alice and Bob, in which payments are sent in both directions (see the code sample below). In this case, both Alice and Bob have the same control over the transaction contents including exchange rates and expiration time.
```

dips/dip-14.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+```
+pub enum AccountAuthenticator {
+    /// Single signature
+    Ed25519 {
+        public_key: Ed25519PublicKey,
+        signature: Ed25519Signature,
+    },
+    /// K-of-N multisignature
```
bob-wilson, line 149: It's probably worth adding some additional explanation above to avoid confusion. This K-of-N multisig is orthogonal to the multi-agent transactions proposed here.

dips/dip-14.md
```diff
@@ -0,0 +1,245 @@
...
+        signature: MultiEd25519Signature,
+    },
+}
+```
+
+
+
+### Add New Variants to `TransactionAuthenticator`
+
+We have added two new variants to `TransactionAuthenticator` to differentiate single-agent and multi-agent transaction scheme. In the single-agent scheme, sender is the only signer and sender’s signature is verified against the `RawTransaction`. While in the multi-agent scheme, the sender and all the secondary signers have to sign over `RawTransactionWithData::MultiAgent{ raw_txn, secondary_signer_addresses }`, to make sure that all parties agree to transact with each other.
```
bob-wilson, line 161: The comments in the code show that the Ed25519 and MultiEd25519 variants are kept for backward compatibility, but I would also suggest a sentence in the text here to explain that those variants are now moved into the new AccountAuthenticator enum above.

dips/dip-14.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+
+
+### Additional Checks During Transaction Validation
+
+In addition to signature checking, we also perform the following additional checks on signed multi-agent transactions during transaction validation:
+
+* The number of secondary authenticators in the signed transaction is the same as the number of secondary signer addresses that all parties have signed over.
+* There are no duplicates in the signers. In other words, sender and all the secondary signers have distinct account addresses.
```
bob-wilson, line 197: Why is this necessary? What bad things could happen if we allowed duplicates? There's a cost to check that, so I assume there's a good reason but offhand it seems fairly harmless.

bob-wilson (commented):
I made a few trivial suggestions (e.g., blank space before parenthesis) and also suggested a few spots where it might help to further clarify the description.

dips/dip-14.md
```diff
@@ -0,0 +1,244 @@
+---
+dip: 14
+title: Multi-Agent Transactions
+author: Emma Zhong(@emmazzz), Tim Zakian, Sam Blackshear
```
xli, line 4: author -> authors. (otherwise, build fails)

emmazzz: Ah that explains a lot.. Thanks!

dips/dip-14.md
emmazzz: This is useful in scenarios like voting: we certainly don't want duplicate signers. Although we can also perform this check on-chain in Move code. 

I think disallowing duplicates in signers can also be quite powerful from verification point of view. Imagine if the code modifies resources stored under signer A's address, with the duplicates checking we have guarantee that this modification does not modify anything owned by signer B.

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+        secondary_signer_addresses: Vec<AccountAddress>,
+    },
+}
+```
+
+
+
+### Add New Enum `AccountAuthenticator`
+
+Previously each transaction can only have one signer, which is the sender of the transaction. In the multi-agent scheme, with the ability to have multiple signers, the transaction can potentially have multiple authenticators from different accounts and of different schemes. An `AccountAuthenticator` serves as the authenticator for one account. And a `TransactionAuthenticator` can contain multiple `AccountAuthenticator`s, as shown in the next subsection.
```
aching, line 140:
```suggestion
Previously each transaction can only have one signer, which is the sender of the transaction. In the multi-agent scheme, with the ability to have multiple signers, the transaction can potentially have multiple authenticators from different accounts and of different schemes. An `AccountAuthenticator` serves as the authenticator for one account. A `TransactionAuthenticator` can contain multiple `AccountAuthenticator`s, as shown in the next subsection.
```

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+    secondary_signer_addresses: vector<address>,
+    secondary_signer_public_key_hashes: vector<vector<u8>>,
+    txn_gas_price: u64,
+    txn_max_gas_units: u64,
+    txn_expiration_time: u64,
+    chain_id: u8,
+)
+```
+ 
+Following are the additional checks it performs.
```
aching, line 216: This is a bit odd, may be reword?  "Below are the additional checks that are performed." or something?

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+}
+```
+
+
+
+### Add New Enum `AccountAuthenticator`
+
+Previously each transaction can only have one signer, which is the sender of the transaction. In the multi-agent scheme, with the ability to have multiple signers, the transaction can potentially have multiple authenticators from different accounts and of different schemes. An `AccountAuthenticator` serves as the authenticator for one account. And a `TransactionAuthenticator` can contain multiple `AccountAuthenticator`s, as shown in the next subsection.
+
+Notice that `AccountAuthenticator` has a multi-signature variant `MultiEd25519`. `MultiEd25519` is a signature scheme an account uses to generate signatures. This is different from multi-agent, which is a new type of transaction.
```
aching, line 142: Does this mean that an account using MultiEd25519 can't be part of a MultiAgentTransaction?

I don't know much about which accounts use MultiEd25519, though.  Is there a reference on that?

aching:
This is really well written and thoughout @emmazzz, thanks!

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+---
+
+
+# Summary
+
+This DIP describes a new scheme of user transaction: multi-agent transactions.
+
+# Abstract
+
+Currently in the Diem Framework, a transaction acts on behalf of a single on-chain account. However, there is no mechanism for multiple on-chain accounts to agree on a single atomic transaction. This DIP presents a new scheme of transactions--multi-agent transactions--which act on behalf of multiple on-chain accounts. Multi-agent transactions leverage Move’s [_`signer`_](https://developers.diem.com/docs/move/move-signer/) type to allow essentially any arbitrary atomic actions in one transaction involving multiple on-chain accounts.
```
moeziniaf, line 18:
```suggestion
Currently in the Diem Framework, a transaction acts on behalf of a single on-chain account. However, there is no mechanism for multiple on-chain accounts to agree on a single atomic transaction. This DIP presents a new scheme of transactions--multi-agent transactions--which act on behalf of multiple on-chain accounts. Multi-agent transactions leverage Move’s [_`signer`_](https://developers.diem.com/docs/move/move-signer/) type to allow an arbitrary number of atomic actions in one single transaction involving multiple on-chain accounts.
```

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+
+# Motivation/Use Cases
+
+## Minting Directly to VASPs
+
+Today in the Diem Framework, we need two transactions in order to mint money through designated dealers to VASPs. The first transaction is a tiered-mint transaction sent by the treasury compliance account to mint money to a designated dealer’s account. The second transaction transfers the minted funds from the designed dealer’s account into the VASP’s account. This procedure requires two transactions due to the restriction that each transaction can only have one signer argument. In the multi-agent scheme, this restriction no longer exists and we can perform both steps in a single atomic transaction (see code sample below). This atomicity implies that designated dealers no longer need custody services for the money they temporarily hold in their accounts, making it more affordable and much quicker to operate as a designated dealer. This should make it much easier for more DDs to onboard in the Diem ecosystem, providing more options for VASPs. It is highly likely that in the future, DDs primarily mint directly on requests from VASPs.
```
moeziniaf, line 29:
```suggestion
Today in the Diem Framework, we need two transactions in order to mint money through designated dealers to VASPs. The first transaction is a transaction sent by the treasury compliance account to mint money into a designated dealer’s account. The second, decoupled transaction transfers the minted funds from the designated dealer’s account into a VASP’s account. This procedure requires two transactions due to the restriction that each transaction can only have one signer argument, despite this being a multi-party flow.

This restriction no longer exists in our proposed multi-agent scheme and we can perform both steps in a single atomic transaction (see code sample below). This atomicity implies that designated dealers no longer need custody services for the money they temporarily hold in their accounts, making it more affordable and much quicker to operate as a designated dealer. This should make it much easier for more DDs to onboard in the Diem ecosystem, providing more options for VASPs. It is highly likely that in the future, DDs primarily mint directly on requests from VASPs.
```

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+   tc: signer,
+   dd: signer,
+   vasp_addr: address,
+   amount: u64,
+   tier_index: u64
+) {
+   // First, TC mints to DD.
+   tiered_mint<Coin>(
+       tc, address_of(dd), amount, tier_index
+   );
```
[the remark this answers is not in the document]
moeziniaf: Yes, I think this code sample is using existing transactions available today (since `tiered_mint` requires the received to be a DD). However, I think for the sake of this proposal let's show that we can do this with a single move function call
`mint<Coin>(tc, dd, vasp_addr, amount)` which would mint directly to vasp.
It's OK to show this conslidation of functions here imo

dips/dip-169.md
```diff
@@ -0,0 +1,261 @@
...
+
+    // Then, Bob pays Alice in currency EUR.
+    // Previously, this line has to be a separate transaction.     
+    pay<EUR>(bob, address_of(alice), amount_eur);
+}
+```
+
+
+
+## Actual Dual-attested Transactions
```
moeziniaf, line 76:
```suggestion
## Dually-attested Onchain Transactions 
```?

File: dips/dip-169.md (hunk 10 above)
Comment (moeziniaf) [reply dips/dip-169.md:142]: +1. I think it would be useful to explain how to reconcile `MultiAgent` and `MultiEd25519` if one is a sig scheme and another a txn scheme and why they are both variants of `TransactionAuthenticator`?

File: dips/dip-169.md [hunk 15]
```diff
@@ -0,0 +1,261 @@
...
+        signer_public_key_hash == *&signer_account.authentication_key,
+        Errors::invalid_argument(PROLOGUE_EINVALID_ACCOUNT_AUTH_KEY),
+    );
+    i = i + 1;
+};
+```
+
+# Feature Gating
+
+Since there is no application of multi-agent transaction scheme in Diem yet, this feature is gated behind `DiemVersion = 10`. Multi-agent transactions are rejected when the Diem version is less than 10.
```
moeziniaf, line 258: why 10?

bob-wilson: I think we should omit the feature gating from the DIP. IMO, this proposal to use version 10 should not have ever been anything more than a temporary hack to let us land the PR while we were figuring out what, if any, kind of feature gating to use for this. That discussion still hasn't settled and it seems beyond the scope of the DIP regardless.

emmazzz: I heard back from @sblackshear that he and @aching have agreed that multi-agent can be gated at Diem version 3. I have updated the DIP to reflect that.

aching: @emmazzz, I would agree with @bob-wilson that it might be hard to make this promise in this DIP about which version it would enabled in.  It might be useful to leave that detail until the "final" phase if we want to include it - although I also not sure whether it really belongs in a DIP at all.

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+# Abstract
+
+Currently in the Diem Framework, a transaction acts on behalf of a single on-chain account. However, there is no mechanism for multiple on-chain accounts to agree on a single atomic transaction. This DIP presents a new scheme of transactions--multi-agent transactions--which act on behalf of multiple on-chain accounts. Multi-agent transactions leverage Move’s [_`signer`_](https://developers.diem.com/docs/move/move-signer/) type to allow an arbitrary number of atomic actions in one single transaction involving multiple on-chain accounts.
+
+## Terminology:
+
+* Primary signer: This is the account that the transaction is sent from. This account’s sequence number is incremented, and is the account that pays gas. There is precisely one of these for every transaction.
+* Secondary signer: This is any account that participates in a multi-agent transaction that isn’t the primary sender. There can be 0-N of these. The 0 case is a normal transaction today. 
+
+# Motivation/Use Cases
```
dahliamalkhi, line 25: all the use-cases/examples who two-agent tx's. could we have an example with (say) three?

dips/dip-169.md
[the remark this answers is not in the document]
dahliamalkhi: +1 from me on @moeziniaf 's suggestion, it is not clear that dd's would need to hold balances on-chain in the future

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+    // Then, Bob pays Alice in currency EUR.
+    // Previously, this line has to be a separate transaction.     
+    pay<EUR>(bob, address_of(alice), amount_eur);
+}
+```
+
+
+
+## Dually-attested Onchain Transactions 
+```?
```
dahliamalkhi, line 79: leftover?

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+
+Some benefits brought by the multi-agent scheme in this scenario are:
+
+* Extensibility: In the current scheme, our dual attestation protocol is very specific to travel rule payments. And extending this protocol to dual attestation on a different type of action would require another ad-hoc data format and signature scheme. The multi-agent transaction scheme generalizes this by allowing dual attestation on any action encodable by a Move script without having to define new data formats or signature schemes.
```
dahliamalkhi, line 85:
```suggestion
* Extensibility: In the current scheme, our dual attestation protocol is very specific to travel rule payments. Extending this protocol to dual attestation on a different type of action would require another data format and signature scheme. The multi-agent transaction scheme generalizes this by allowing dual attestation on any action encodable by a Move script without having to define new data formats or signature schemes.
```

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+* Security: Multi-agent scheme closes two security loopholes of our current protocol:
+    * Replayability: Since the payee only signs over the payer and the amount of the payment, today the payer can replay the transaction using old payee approvals. In the new scheme, this wouldn’t be possible since the payee has to sign over the sequence number of the transaction as well, disallowing any replay of the transaction.
```
dahliamalkhi, line 87: I would be a bit cautious with this statement. Isn't there protection against this? Isn't the receiver signing the full transaction which includes a unique sequence number, and if replayed by the sender, it will be rejected?

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+    * Currency: Today’s protocol does not let the payee commit to the currency type so the payer can reuse a payee approval for payment in one currency for payment in another with the same amount. Again, since the payee has to sign over the entire transaction including the currency type, this scenario would be impossible with the multi-agent scheme.
```
dahliamalkhi, line 88: same as above.

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+    pay<EUR>(bob, address_of(alice), amount_eur);
+}
+```
+
+
+
+## Dually-attested Onchain Transactions 
+```?
+
+Currently, we execute a [_dual-attested payment_](https://dip.diem.com/dip-1/) transaction by including a signature of the payee as an argument to the transaction, and checking the validity of the signature on-chain if the payment is between two different VASPs and the amount is over a certain threshold. In order to accomplish that, we have to run some fairly complex code to reconstruct and verify the signature ad-hoc on-chain. With the new scheme, we can simply require that the payee signs the entire transaction as well, and verify the signatures of both the payer and payee while validating the transaction. The code samples below show the before and after of what such a transaction might look like.
```
dahliamalkhi, line 81: the "fairly complex code to reconstruct and verify" below doesn't look very complex. Perhaps instead say: "Currently ... we need to reconstruct and verify the signature on chain".

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+# Code Changes to Support Multi-agent
+
+## DiemVM Code Changes
+
+### Add New Enum `RawTransactionWithData`
+
+In order to preserve backward compatibility, we can’t add more fields representing secondary signers to `RawTransaction` struct. Thus we add the following additional enum `RawTransactionWithData`, which the sender and secondary signers will sign over to authenticate the transaction.
```
dahliamalkhi, line 127:
```suggestion
In order to preserve backward compatibility, we can’t add more fields representing secondary signers to `RawTransaction` struct. Thus we wrap `RawTransaction` in an enum `RawTransactionWithData`, which the sender and secondary signers will sign over to authenticate the transaction.
```

dips/dip-169.md
dahliamalkhi: I believe the difference is that MultiAgent does not require to combine signatures off-chain, and the combined signature is a single signature corresponding to a single account

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+    MultiEd25519 {
+        public_key: MultiEd25519PublicKey,
+        signature: MultiEd25519Signature,
+    },
+}
+```
+
+
+
+### Add a New Variant to `TransactionAuthenticator`
```
dahliamalkhi, line 164: Add a New Variant or a new field?

dahliamalkhi (changes requested):
very cool proposal.

dips/dip-169.md
sblackshear: There is no protection against this. The recipient does not sign the full transaction, only an `(amount, sender_address, metdata)` approval message. This approval message can be replayed an arbitrary number of times.

dips/dip-169.md
sblackshear: The approval message does not contain the currency, so this is indeed also a problem with the current scheme.

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+){
+    pay<Coin>(payer, address_of(payee), amount, metadata);
+}
+```
+
+
+
+## Arbitrary Atomic Actions in One Transaction
+
+Besides the use cases mentioned above, our new multi-agent transaction scheme allows the sender to encode any arbitrary combination of actions in a single atomic transaction. This added expressiveness opens up a lot more possibilities in the type of scenarios that we can implement. Some examples include [_delivery versus payment_](https://www.investopedia.com/terms/d/dvp.asp) (a crucial feature for securities settlement) and atomic administrative approval of sensitive actions (e.g., account creation).
```
sblackshear, line 116:
```suggestion
Besides the use cases mentioned above, our new multi-agent transaction scheme allows the sender to encode any combination of actions in a single atomic transaction. This added expressiveness opens up a lot more possibilities in the type of scenarios that we can implement. Some examples include [_delivery versus payment_](https://www.investopedia.com/terms/d/dvp.asp) (a crucial feature for securities settlement) and atomic administrative approval of sensitive actions (e.g., account creation). 
```

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+## Arbitrary Atomic Actions in One Transaction
+
+Besides the use cases mentioned above, our new multi-agent transaction scheme allows the sender to encode any arbitrary combination of actions in a single atomic transaction. This added expressiveness opens up a lot more possibilities in the type of scenarios that we can implement. Some examples include [_delivery versus payment_](https://www.investopedia.com/terms/d/dvp.asp) (a crucial feature for securities settlement) and atomic administrative approval of sensitive actions (e.g., account creation). 
+
+
+
+
+# Code Changes to Support Multi-agent
+
+## DiemVM Code Changes
```
sblackshear, line 123:
```suggestion
## Diem Adapter Code Changes
```

dips/dip-169.md
sblackshear, line 127:
```suggestion
To preserve backward compatibility, we can’t add more fields representing secondary signers to `RawTransaction` struct. Thus we add the following additional enum `RawTransactionWithData`, which the primary and secondary signers will sign over to authenticate the transaction.
```

dips/dip-169.md
sblackshear: - https://github.com/diem/dip/blob/main/dips/dip-11.md#authentication-key for more on MultiEd25519. Any account can use this feature
- Ac account using MultiEd25519 *can* be part of a MultiAgentTransaction (at least based on my reading of Emma's code in https://github.com/diem/diem/pull/7427). But I think we should be careful about opening DDOS attack vectors with this. We already cap the max number of sigs in MultiEd25519 to 32 to prevent an attacker from having too many "free" signature checks. But if we also allow a max of 32 signers in a multi-agent transaction, that's 32 * 32 free signature checks... 
 
Ideally, the addition of the multi-agent feature wouldn't give an attacker any more DDOS power, but I think we will need to tweak the rules/code a bit to ensure that.

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+        signature: Ed25519Signature,
+    },
+    /// K-of-N multisignature
+    MultiEd25519 {
+        public_key: MultiEd25519PublicKey,
+        signature: MultiEd25519Signature,
+    },
+    /// Multi-agent transaction.
+    MultiAgent {
+        sender: AccountAuthenticator,
```
sblackshear, line 183: Terminology nit: we call this "primary signer" at the beginning of the doc. We should probably use that term everywhere instead of "sender", or use sender instead of primary signer? I personally like primary signer because its connection to secondary signers is clearer + it's less overloaded than sender, but consistency is the most important thing here.

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+
+
+### Additional Checks During Transaction Validation
+
+In addition to signature checking, we also perform the following additional checks on signed multi-agent transactions during transaction validation:
+
+* The number of secondary authenticators in the signed transaction is the same as the number of secondary signer addresses that all parties have signed over.
+* There are no duplicates in the signers. In other words, sender and all the secondary signers have distinct account addresses.
```
sblackshear, line 197: It might be nice to comment on what this buys you at the Move level in the Diem Framework: every `signer` value is guaranteed to be backed by a distinct address.

dips/dip-169.md
sblackshear: "Variant" is the right technical term for the one of potential values of a sum type like `TransactionAuthenticator`.

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+
+## Diem Framework Changes
```
sblackshear, line 202: I think describing these changes at a high level instead of including code might make sense. We wouldn't want to have go to back and change the DIP if we make small changes to the code that are unrelated to the multi-agent functionality (e.g., add/remove param from `multi_agent_script_prologue`).

sblackshear:
@emmazzz, one question for you: does multi-agent imply any changes to the "script signature requirements" [here]( https://github.com/diem/dip/pull/153)? I can update that DIP proposal if so.

dips/dip-169.md
bob-wilson: In the context of Diem, I think the better terminology is "Move Adapter". That's what we use in the specification document (`diem/specifications/move_adapter/README.md`).

dips/dip-169.md
```diff
@@ -0,0 +1,263 @@
...
+This DIP describes a new scheme of user transaction: multi-agent transactions.
+
+# Abstract
+
+Currently in the Diem Framework, a transaction acts on behalf of a single on-chain account. However, there is no mechanism for multiple on-chain accounts to agree on a single atomic transaction. This DIP presents a new scheme of transactions--multi-agent transactions--which act on behalf of multiple on-chain accounts. Multi-agent transactions leverage Move’s [_`signer`_](https://developers.diem.com/docs/move/move-signer/) type to allow an arbitrary number of atomic actions in one single transaction involving multiple on-chain accounts.
+
+## Terminology:
+
+* Primary signer: This is the account that the transaction is sent from. This account’s sequence number is incremented, and is the account that pays gas. There is precisely one of these for every transaction.
+* Secondary signer: This is any account that participates in a multi-agent transaction that isn’t the primary sender. There can be 0-N of these. The 0 case is a normal transaction today.
```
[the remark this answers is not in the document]
emmazzz: We don't put a max on `N` here. Yes we should put a limit same as for multisig. Will work on that.

dips/dip-169.md
emmazzz: Good idea. Using "primary signer" would be more clear.

dips/dip-169.md
dahliamalkhi: I see. This is not an issue with single-agent transactions, it is a problem with DIP-1 which could be fixed there, right? I would prefer to tone this down and mention the issue with DIP-1.  

Here, how about: 
The Multi-agent natively leads to robust and secure dual-attestation protocols, since both parties sign the entire transaction.

dips/dip-169.md
dahliamalkhi: same

dahliamalkhi (changes requested):

dips/dip-169.md
sblackshear: I'm fine with mentioning the motivation in more general terms as you suggest, although I do think a specific example helps.

As an aside, it's not easy to fix the problem in DIP1 without this feature + CRSN's. If the approval message signs (e.g.) the payer's sequence number, the payer has difficulty sending other transactions concurrently with a TR transaction without invaliding the payee's approval. And it's not easy to sign over the currency without this feature because currency is a Move type, and Move doesn't allow you to reflect a type into a value that can then been signed.

dips/dip-169.md
emmazzz: A check has been added to allow no more than 32 signatures. DIP has also been updated to talk about this check in section `Additional Checks During Transaction Validation`.

dahliamalkhi (approved):
LGTM

PR #163, opened 2021-04-08
DIP-160 Draft: off-chain API ping command

xli:


dips/dip-160.md
```diff
@@ -0,0 +1,59 @@
...
+
+This DIP describes an off-chain API v2 ping command that makes it easier for clients to detect network and protocol problems.
+
+# Abstract
+
+To improve the ability of off-chain service handling network problems and protocol errors, we’d like to introduce a ping command for client to:
+1. detect dead connections to a remote off-chain service.
+2. measure the latency to a remote off-chain service.
+3. detect off-chain protocol errors to a remote off-chain service, for example: when a remote off-chain service rotates their compliance key, and client used old compliance public key to verify the response message, the ping command can be used to debug the error and verify the fix.
+
```
davidiw, line 24: nit extra newline

dips/dip-160.md
```diff
@@ -0,0 +1,59 @@
...
+# Summary
+
+This DIP describes an off-chain API v2 ping command that makes it easier for clients to detect network and protocol problems.
+
+# Abstract
+
+To improve the ability of off-chain service handling network problems and protocol errors, we’d like to introduce a ping command for client to:
+1. detect dead connections to a remote off-chain service.
+2. measure the latency to a remote off-chain service.
+3. detect off-chain protocol errors to a remote off-chain service, for example: when a remote off-chain service rotates their compliance key, and client used old compliance public key to verify the response message, the ping command can be used to debug the error and verify the fix.
```
davidiw, line 23: maybe simplify this last one. For example, an off-chain service responds back with an unexpected compliance key due to key rotation, this can be leveraged to debug and verify the inconsistency.

dips/dip-160.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+# Specification
+
+The PingCommand object definition:
```
davidiw, line 28: object's ?

dips/dip-160.md
```diff
@@ -0,0 +1,59 @@
...
+And the response of the `PingCommand`:
+```
+{
+    "_ObjectType": "CommandResponseObject",
+    “status”: “success”,
+    "cid": "12ce83f6-6d18-0d6e-08b6-c00fdbbf085a"
+}
+```
+
+Note we don’t need a random message body in PingCommand for the server to response, because the off-chain API protocol already have a `cid` field defined in the `CommandRequestObject` and `CommandResponseObject` to serve the purpose of distinguish different command instance.
```
davidiw, line 57: Note, PingCommand does not require a random payload, as the Off-Chain Protocol already has a `cid` to uniquely distinguish different `CommandRequestObject`s and `CommandResponseObject`s.

dips/dip-160.md
```diff
@@ -0,0 +1,59 @@
...
+
+There is no `command_error` for this command. Server may respond with `protocol_error` (with CommandResponseObject.status==“failure”) defined in [DIP-1](https://dip.diem.com/dip-1) when processing an invalid HTTP headers, JWS message or `CommandRequestObject` payload.
```
davidiw, line 59: I would probably just leave it as "This protocol introduces no new errors outside the scope of DIP-1."

davidiw (changes requested):
hit me up after tweaks and we should get this landed

davidiw (approved):
I think this is actually a standard type, but no one else is using it right, so :shrug:

PR #167, opened 2021-04-20
[DIP-20] Address last call phase comments

tzakian:
Addresses comments from the last call phase. Mostly minor changes.

### DIFF
--- a/dips/dip-20.md
+++ b/dips/dip-20.md
@@ -7,6 +7,10 @@ type: Informational
 created: 08/28/2020
 ---
 
+### DISCLAIMER:
+
+The functionality described in this DIP (including but not limited to, XDX) may not be available on DPN at the time of publication of this DIP.
+
 ---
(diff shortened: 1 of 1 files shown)


dips/dip-20.md
```diff
@@ -187,7 +191,7 @@ denomination coin in common usage is one cent (1/100th of an XUS dollar),
 so the `fractional_part` for XUS would be
 100. If you wanted to translate from an on-chain value `v` to a number that
```
bob-wilson, line 192: At least from what I can see in this diff, there is no subsequent reference to the value `v`, so there should be no reason to specify that name. This can be just "...from an on-chain value to a number..."

bob-wilson (approved):
These changes all look good to me. I noticed one little nit unrelated to this PR.

dahliamalkhi (approved):
I think we can "Finalize" this!

PR #170, opened 2021-05-07
[DIP-168] Initial draft of DIP on CRSNs

tzakian:
This PR is the initial draft of the proposal for Conflict-Resistant Sequence Numbers (CRSN)s for Diem.


dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+
+An account `A` that has opted-in to using a CRSN may force expire either a single sequence nonce, or may expire a batch of sequence nonces. A sequence nonce that has been expired will cause any transactions that have been assigned that sequence nonce to be rejected according to the Acceptance and Rejection Criteria above.
+
+A singular sequence nonce may be expired by a no-op transaction assigned the sequence nonce that is to be expired. This will then perform the actions described in 3(a). Any further transactions assigned this sequence nonce will then be rejected by the Acceptance and Rejection Criteria.
+
+If a range of sequence nonces is to be expired, it can be done so by sending a specific force-expiration transaction — `force_shift(j)` — that contains the amount by which the window should be shifted `j`. This can be informally thought of as incrementing the account's `CRSN(k).min_nonce`  by `j`, and performing a "left shift" on the account's `CRSN(k).slots. `More precisely: let `CRSN(k)` be `A`'s CRSN at the time of processing the `force_shift(j)` transaction and let this transaction be sent with sequence none `m`, then the following updates will be performed (in order):
```
aching, line 86: IIUC, a single no-op txn that invalidates more than one sequence number will break the existing sequence number property that every account, sequence number maps to a single transaction.

I get this would be more efficient rather than sending 1 txn for every shift, but I don't know what downstream effects this could have.  Is this a concern?

dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Transactions should first be ordered by gas price and sending account role as they are today
+2. Amongst equal transactions according to the ordering in (1) ties should be broken for smaller sequence nonces greater than the sending account's current `min_nonce` .
+3. Transaction's with a sequence nonce less than the sending account's current `min_nonce` should be cleared from the mempool. A transaction can only enter the state where it needs to be cleared due to force expiration.
+
+# CRSNs Satisfy Requirements
+
+1. **Non-Replayability:** If two transactions are sent from the same account `A` with the same sequence nonce `i` then the latter will be rejected by (3b), or `i` will be less than `A.CRSN(k).min_nonce` and will be rejected by (2).
+2. **Concurrency:** Transactions from the same account can be processed concurrently as long as their slot is open in the current state `A.CRSN(k)`.
+3. **Force Expiration:** A transaction with sequence nonce `i` can only be expired by processing another transaction with the same sequence nonce. A transaction can't be accidentally expired since the window may only move over filled slots (i.e., processed transactions).
```
aching, line 108: Does the force_shift(j) violate this property?

aching (commented):
Sorry for the delay on this, great proposal @tzakian !

dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+
+The acceptance and rejection criteria for a pair `(i, CRSN(k))` where `i` is the sequence nonce assigned to a transaction sent by `A` and `CRSN(k)` is the CRSN state under `A` at the time of processing is as follows:
+
+1. If `i` is greater than or equal to `CRSN(k).min_nonce + k` the transaction    is discarded.
+2. If `i` is less than `CRSN(k).min_nonce` the transaction is discarded.
+3. If `i` is greater than or equal to `CRSN(k).min_nonce` and less than    `CRSN(k).min_nonce + k` then
+    1.  If the slot at position `i` in `CRSN(k).slots` is `false` then `CRSN(k).slots[i]` is set to `true` and the transaction is accepted. Subsequently, the [shifting algorithm](#shifting-algorithm) is applied to `CRSN(k)`. `A`'s sequence number is incremented.
+    2. If the slot at position `i` in `CRSN(k).slots` is `true` the transaction is rejected.
+
+## Shifting algorithm
```
aching, line 66: I love the math definition, but some graphics might help to illustrate this if you have some.

dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+
+Let `j = max {m | CRSN(k).min_nonce <= e <= m < k . forall CRSN(k).slots[e] == true }` (i.e., the least index in `CRSN(k).slots` such that all slots less than it are set to `true`).  The following updates are then performed:
+
+* `CRSN(k).min_nonce = CRSN(k).min_nonce + j`
+* for all `j <= i < k, ``CRSN(k).slot[i - j] = CRSN(k).slot[i]` and, for all `k - j <= i < k, ``CRSN(k).slot[i] = false`.
+
+## Sequence number to CRSN transition
```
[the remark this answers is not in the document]
tzakian: Good question. Sam and I discussed this, and while an account could transition back, the account's sequence number  when transitioning from CRSN -> seqno would have to be set at `CRSN(k).min_nonce + k` to be safe. This is because the account could have processed transactions with sequence nonces anywhere in the range of `(CRSN(k).min_nonce, CRSN(k).min_nonce + k)` and if we didn't do this it could allow transaction replay for transactions in this window.

This however causes issues if we want to keep sequence numbers sequential, because this would cause a jump in the sequence number at the time of transition.

dips/dip-168.md
tzakian: Good callout. I'll clarify here that "... with the same sequence nonce, or by a `force_shift` operation is performed.", and "... the window may only move over filled slot (i.e., processed transaction) unless a specific `force_shift` operation is performed by the sender."

dips/dip-168.md
tzakian: I do have some! Will add :)

dips/dip-168.md
tzakian: Good question. The `force_shift` operation would be a transaction itself, and it's important to note here (and I'll clarify this in this paragraph) that this force shift does not apply to the sequence number -- the `force_shift` transaction will still cause the account's sequence number to increase by one, and that's it. Because of this, it will still maintain the mapping of (account, sequence_number) mapping to a single transaction.

dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+Each transaction sent from `A` is assigned a number or *sequence nonce* `i: u64`. This number coupled with the state `CRSN(k)` held under `A` defines the acceptance and rejection criteria for transactions.
+
+## Acceptance and Rejection Criteria
+
+The acceptance and rejection criteria for a pair `(i, CRSN(k))` where `i` is the sequence nonce assigned to a transaction sent by `A` and `CRSN(k)` is the CRSN state under `A` at the time of processing is as follows:
+
+1. If `i` is greater than or equal to `CRSN(k).min_nonce + k` the transaction    is discarded.
+2. If `i` is less than `CRSN(k).min_nonce` the transaction is discarded.
+3. If `i` is greater than or equal to `CRSN(k).min_nonce` and less than    `CRSN(k).min_nonce + k` then
+    1.  If the slot at position `i` in `CRSN(k).slots` is `false` then `CRSN(k).slots[i]` is set to `true` and the transaction is accepted. Subsequently, the [shifting algorithm](#shifting-algorithm) is applied to `CRSN(k)`. `A`'s sequence number is incremented.
```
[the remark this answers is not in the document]
tzakian: The account `A` will still maintain its sequence number. It will be incremented for every transaction that is sent, but the transaction will not contain the sequence number.

This will maintain the invariant that for all accounts (CRSN or not) that `A.sequence_number == <number of transactions sent by A that have been processed>`. I'll add this in.

dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+
+### High-Frequency Transactions
+
+If an account is issuing a high number of transactions, blocking on other transactions to process in order to send/process their transaction before sending another presents a number of issues if, e.g., the sender's wallet is sharded or distributed across multiple different nodes. Purely sequential sequence numbers prohibit concurrently processing multiple transactions in this setting and recovering from any one transaction being possibly rejected--since future transactions that have already been sent to the network will block until a new transaction has been sent and processed by the network.
```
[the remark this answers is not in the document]
tzakian: ah! accidentally slipped past in the editing process. Will remove.

dips/dip-168.md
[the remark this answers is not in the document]
aching: Yeah, that seems to violate the "A.sequence_number == <number of transactions sent by A that have been processed>"

dips/dip-168.md
```diff
@@ -0,0 +1,114 @@
...
+
+It is important to note that the sequence nonce assigned to the `force_shift(j)` transaction must be accepted by the Acceptance and Rejection Criteria in the pre-state in order for these changes to be performed.
+
+## Transaction Prioritization & Validation
+
+A transaction will only be allowed into mempool if the transaction's sequence nonce is greater than or equal to the sending account's current `min_nonce`. When validating a transaction immediately before execution, the transaction will be discarded if the sequence nonce provided with the transaction is not accepted according to the Acceptance and Rejection Criteria. With exception of the addition of the CRSN logic and checks determining if the account is using a sequence number or CRSN, transaction validation both when entering the mempool, and before execution is unchanged.
+
+After a transaction using a CRSN has entered mempool, there must be a prioritization mechanism for it when filling a block. Note that while the prioritization criteria are not part of the specification of CRSNs, the following amended rules is the most likely of these for accounts that have opted-in to using a CRSN:
```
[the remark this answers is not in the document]
tzakian: Very good point. I have a PR up to add support to mempool here: https://github.com/diem/diem/pull/8528 if you want to see how I've implemented them. I can either writeup a smaller DIP for this, or roll it into this DIP. I think I have a preference for the latter, but happy to write it up separately if you think that would be better.

The current idea is to keep the same number for both CRSN and seqno accounts, however the criteria of how to remove transactions from the mempool needs to differ: in CRSNs only transactions that are less than the `min_nonce` field under the sending account can be safely removed from mempool (as they're the only ones we know for sure cannot be processed).

tzakian:
I believe I've addressed all of the comments. Let me know if I missed anything!

dips/dip-168.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Acceptance and Rejection Criteria
+
+The acceptance and rejection criteria for a pair `(i, CRSN(k))` where `i` is the sequence nonce assigned to a transaction sent by `A` and `CRSN(k)` is the CRSN state under `A` at the time of processing is as follows:
+
+1. If `i` is greater than or equal to `CRSN(k).min_nonce + k` the transaction    is discarded.
+2. If `i` is less than `CRSN(k).min_nonce` the transaction is discarded.
+3. If `i` is greater than or equal to `CRSN(k).min_nonce` and less than    `CRSN(k).min_nonce + k` then
+    1.  If the slot at position `i` in `CRSN(k).slots` is `false` then `CRSN(k).slots[i]` is set to `true` and the transaction is accepted. Subsequently, the [shifting algorithm](#shifting-algorithm) is applied to `CRSN(k)`. `A`'s sequence number is incremented.
+    2. If the slot at position `i` in `CRSN(k).slots` is `true` the transaction is rejected.
```
davidiw, line 64: don't rejects cost gas? is the assumption that a validator mempool would note this as already committed and discard it?

dips/dip-168.md
```diff
@@ -0,0 +1,124 @@
...
+
+1. `CRSN(k).slots[m]` is set to `true`.
+2. `CRSN(k).min_nonce = CRSN(k).min_nonce + j`
+3. for all `j <= i < k, ``CRSN(k).slot[i - j] = CRSN(k).slot[i]` and, for all `k - j <= i < k, ``CRSN(k).slot[i] = false`.
+
+It is important to note that the sequence nonce assigned to the `force_shift(j)` transaction must be accepted by the Acceptance and Rejection Criteria in the pre-state in order for these changes to be performed. This does not affect the account's sequence number, and the `force_shift` transaction will increment the sending account's sequence number as defined in the Acceptance and Rejection Criteria.
+
+## Transaction Prioritization & Validation
```
davidiw, line 104: @gregnazario for mempool CRSN

davidiw (commented):
solid doc, easy to follow

dips/dip-168.md
aching: Reject is gas free and can happen in a mempool or during execution.  Aborts are the transactions that pay gas and appear on-chain.

davidiw: ah, is it discard or reject, 🤦

aching (approved):
Awesome draft!

PR #173, opened 2021-05-15
Initial draft of Dip 172: P2M On-Chain Registration

dimroc:
Write up of the Customer VASP registration step, including on-chain structs and events.

### DIFF
--- a/dips/dip-172.md
+++ b/dips/dip-172.md
@@ -1,0 +1,70 @@
+---
+dip: 172
+title: P2M (eCommerce) On-Chain Registration
+authors: Dimitri Roche (@dimroc), David Wolinsky (@davidw), Yinon Zohar, Daniel Prinz
+status: Draft
+type: Standard
+created: 05/14/2021
+---
+
+# Summary
+
+This DIP describes the registration step for VASPs to enable eCommerce peer to merchant (P2M) payments. By writing on-chain a VASP’s P2M configuration, future off-chain P2M VASP activity will have the information needed to complete the payment process.
+
+This DIP builds on previous work in the space of P2M and identity. [DIP-158](https://github.com/diem/dip/blob/main/dips/dip-158.md) describes one approach for supporting P2M leveraging the Diem off-chain protocol defined in DIP-1. DIP-10 defines an on-chain naming service for VASPs, namely the domain portion of a DiemId.
+
+While there might be many P2M flows, three basic elements can be used to construct many of them:
+A flag signifying whether or not the VASP supports the customer flows for P2M payments
+An authentication URL to kick start the peer authentication and authorization within the customer’s VASP of [DIP-158](https://github.com/diem/dip/blob/main/dips/dip-158.md)
+An URL that links to an image representing the VASPs business
+
+---
+
+# Example Flow
+
+The scenario begins with a customer, Alice, entering the checkout flow of a merchant website:
+1. Alice is presented with several options for payment, one of which is “Pay with Diem”.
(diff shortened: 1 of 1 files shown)


dimroc:
cc @YinonFirstDAG, @danprinz

dips/dip-172.md
```diff
@@ -0,0 +1,70 @@
+---
+dip: 172
+title: P2M (eCommerce) On-Chain Registration
+authors: Dimitri Roche (@dimroc), David Wolinsky (@davidw), Yinon Zohar, Daniel Prinz
```
davidiw, line 4: add everyone's github handle, we should be consistent about that

dips/dip-172.md
```diff
@@ -0,0 +1,70 @@
...
+status: Draft
+type: Standard
+created: 05/14/2021
```
davidiw, line 7: add updated: https://dip.diem.com/dip-0/#updated-header

dips/dip-172.md
```diff
@@ -0,0 +1,70 @@
...
+1. Alice is presented with several options for payment, one of which is “Pay with Diem”.
+1. Upon selecting that option, Alice is presented with a typeahead to enter their DiemId, e.g., alice@CustomerVASP.
+1. Simultaneously the merchant site initiates [DIP-158](https://github.com/diem/dip/blob/main/dips/dip-158.md) with its VASP, the merchant VASP, which returns to the merchant site a reference id that represents the payment agreement and the merchant VASP’s on-chain account address.
+1. The merchant site leverages the on-chain configuration to translate CustomerVASP into an authentication url for iframe integration and includes the reference id and account address provided earlier.
+1. Alice then authenticates with her VASP (CustomerVASP) and confirms the payment agreement with the merchant resulting in the completion of the [DIP-158](https://github.com/diem/dip/blob/main/dips/dip-158.md) protocol.
+
+
+![OnChain Customer VASP Registration](https://static.swimlanes.io/c359dbf4495430287bfc49a66a613176.png)
+
+![P2M Customer Redirect Flow](https://static.swimlanes.io/655c351bab8eef00d30a4e33a3818f24.png)
```
davidiw, line 35: add these to the static file once they stabilize and look at DIP-1 for how to embed an image though I think its fine if we just point to the entry in the github repo.

dips/dip-172.md
```diff
@@ -0,0 +1,70 @@
...
+    enabled: bool,
+    authentication_url: vector<u8>, // UTF-8 encoded string
+    image_url: vector<u8>, // UTF-8 encoded string
+}
+```
+
+#### Field definitions:
+* `enabled` -- defines whether or not the VASP is currently supporting P2M checkout. The intent here is that the customer experience should be seamless and not require awkward checks between the merchant VASP and customer VASP to validate whether or not the customer VASP supports merchant checkout. If this is false or the CustomerVASPConfig does not exist, then the customer VASP does not support checkout.
+* `authentication_url` -- an endpoint that provides an authentication and authorization flow for the customer, the expected format is defined in [DIP-158’s appendix](https://github.com/diem/dip/blob/main/dips/dip-158.md#appendix-a---prerequisite-sharing-common-payment-identifier-and-address)
+* `image_url` -- an endpoint that points to a standard image type (png, jpeg, gif) and is expected to be no larger than 100 KB (or maybe 10 KB?, the merchant or merchant VASP can always resize larger images to be smaller).
```
davidiw, line 54: might want to move discussions like this to the issue (what size should we recommend)

davidiw (approved):
Gonna land this as is. Let's move outstanding questions to the issue so we can follow and close them.

PR #177, opened 2021-05-27
DIP-158: Follow ups

YinonFirstDAG:
Comment about error codes (just a recommendation)
Change sequences to QR code
Update typos


YinonFirstDAG:
Opened by error

PR #185, opened 2021-06-15
DiemID Refactor

sunmilee:
Refactoring DiemID DIP into three DIPS: Diem ID (DIP-10), On-chain domain resolution (https://github.com/diem/dip/pull/187), and Reference ID exchange (https://github.com/diem/dip/pull/186).

### DIFF
--- a/dips/dip-10.md
+++ b/dips/dip-10.md
@@ -1,11 +1,11 @@
 ---
 dip: 10
-title: DiemID Spec
+title: DiemID
 authors: Sunmi Lee (@sunmilee), David Wolinsky (@davidiw), Andrey Chursin(@andll), Kevin Hurley (@kphfb)
 status: Draft
 type: Informational
 created: 11/03/2020
-last_updated: 04/28/2021
+last_updated: 06/15/2021
 issue: https://github.com/diem/dip/issues/156
 ---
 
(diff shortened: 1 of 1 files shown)


davidiw (approved):
Should we call it sender's VASP or sender VASP?

dips/dip-10.md
```diff
@@ -51,75 +51,26 @@ Example: `alice@avasp`
...
+Once the sender VASP gets the receiving VASP's address using on-chain domain lookup, the sender VASP initiates a reference ID exchange as defined in [DIP-183](https://github.com/diem/dip/blob/main/dips/dip-183.md)
+in order settle on a unique reference ID and submit the transaction on-chain.
 
 The format of the command is:
 
 ```
 {
    "_ObjectType": "CommandRequestObject",
-    "command_type": "ReferenceIDCommand",
+    "command_type": "SenderInitiatedReferenceID",
```
xli, line 73: If we are going to rename this command, I suggest new name "ExchangePaymentAddress".
The reason is, for this command, what really need is for sender to get receiver's account address, when sender got receiver's DiemID.
We originally thought reference id is the data need to be exchanged. But I think reference id is an id for referencing the payment that uses DiemID and related account addresses.
It is better to not assume this is a command initiate the reference id, because if we had another command that needs exchange information between VASPs, and the command happens before this command, then we may have trouble.

davidiw: Yeah, naming could definitely be improved!

sunmilee: We can call it SenderInitiatedDiemIdExchange if we want to be specific to be diemID and not mention reference ID, since we're worried that reference ID exchange could happen in other commands

xli: I like this name `SenderInitiatedDiemIDExchange`

davidiw (approved):
good for me

Issue #183, opened 2021-06-09
Reference ID Exchange


sunmilee:

davidiw:
DIP-183 needs to be updated with this feedback:

Considering the general direction here, I'd actually propose a bit of rework in DIP-183 in particular with examples and test data.

Examples which could go into the appendix include various different strategies for exchanging a reference id:

Subaddresses
PayString
Other naming services
Leveraging Diem domain identifiers in addition to other services
Then we could avoid introducing a new naming construct that is Diem specific and altogether unnecessary.

In addition, we should have good test cases for verifying correctness of the exchange.

PR #186, opened 2021-06-15
Sender Initiated Reference ID Exchange

sunmilee:
Reference ID off-chain exchange DIP. Used by DIP-10 Diem ID protocol.

https://github.com/diem/dip/issues/183

### DIFF
--- a/dips/dip-183.md
+++ b/dips/dip-183.md
@@ -13,9 +13,9 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 
 # Summary
 
-This DIP describes a protocol standard for sender initialized pre-flight exchange using reference ID to form payment transactions.
+This DIP describes a protocol standard for sender initialized pre-flight exchange using reference IDs to form payment transactions.
 
-Reference ID is used as a unique identifier for a payment transaction between two parties across two VASPs without revealing any information about those parties on-chain.
+A reference ID is used as a unique identifier for a payment transaction between two parties across two VASPs without revealing any information about those parties on-chain.
 In this protocol, the sender already knows the recipient VASP address and an identifier for the recipient at their VASP. The sender VASP contacts the recipient VASP and shares the sender's identifier, a reference ID, and the recipient's identifier. 
 Upon conclusion, the sender VASP transmits to the blockchain a transaction with the reference ID, completing the protocol without revealing any potentially identifiable user information publicly on-chain.
 
@@ -25,9 +25,9 @@ The sender VASP initiates the off-chain preflight with the following [DIP-1](htt
 ```
 {
    "_ObjectType": "CommandRequestObject",
-    "command_type": "SenderInitiatedReferenceID",
+    "command_type": "SenderInitiatedReferenceIDExchange",
     "command": {
-	    "_ObjectType": "SenderInitiatedReferenceID",
+	    "_ObjectType": "SenderInitiatedReferenceIDExchange",
 	    "sender": "alice",
 	    "sender_address": "dm1pptdxvfjck4jyw3rkfnm2mnd2t5qqqqqqqqqqqqq305frg",
 	    "receiver": "bob",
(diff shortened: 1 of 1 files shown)


dips/dip-183.md
```diff
@@ -0,0 +1,105 @@
...
+## On-Chain Transaction Settlement
+If the amount is below the travel rule limit, the sending VASP can send a p2p transaction with PaymentMetadata and the `reference_id` VASPs agreed on to settle the transaction.
+```
+enum PaymentMetadata {
+    PaymentMetadataV0(ReferenceId),
+}
+type ReferenceId = [u8, 16];
+```
+
+If the amount is at or above the travel rule limit, VASPs should trigger an off-chain protocol for KYC exchange. The same `reference_id` must be used to perform a travel rule exchange as described in [DIP-1](https://github.com/diem/dip/blob/master/dips/dip-1.md).
```
davidiw, line 105:
```suggestion
If the amount is at or above the travel rule limit, VASPs must perform an off-chain identity exchange. The same `reference_id` must be used to perform a travel rule exchange as described in [DIP-1](https://github.com/diem/dip/blob/master/dips/dip-1.md).
```

dips/dip-183.md
```diff
@@ -0,0 +1,100 @@
...
+issue: https://github.com/diem/dip/issues/183
+---
+
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+# Summary
+
+This DIP describes a protocol standard for sender initialized pre-flight exchange using reference ID to form payment transactions.
+
+Reference ID is used as a unique identifier for a payment transaction between two parties across two VASPs without revealing any information about those parties on-chain.
```
aching, line 18: IIUC, each VASP will need to maintain every unique (reference id, sending VASP, receiving VASP) identifier to ensure avoiding duplicate reference ids?

Can VASP A initiate a reference ID R to VASP B
and VASP B also legally initiate a reference ID R to VASP A or is it unique per pair?

sunmilee: Unique per pair, so between VASP A and B, if a reference ID R was used in a previous transaction, regardless of which one initiated it, R cannot be reused for any future transactions between A and B

aching: Thanks for the clarification.  I think it would be good to add something like your statement into the DIP to make this clear.

We should consider from an implementor perspective how feasible this is to implement.  I guess it is expected that the implementor would need to maintain an index of (canonical(VASP1, VASP2), reference ID) and rebuild it from the event history if they lost the index.

This might be simpler if, say, there was just a monotonically increasing reference ID number so one could just look at the last number between canonical(VASP1, VASP, last reference ID number) event on-chain rather than keeping an index?

sunmilee: I think it depends on how the VASP decides to implement it. They can always enforce uniqueness across all reference IDs instead of pair, which is what I do in my miniwallet diem ID example.

dips/dip-183.md
```diff
@@ -0,0 +1,103 @@
...
+
+A reference ID is used as a unique identifier for a payment transaction between two parties across two VASPs without revealing any information about those parties on-chain.
+In this protocol, the sender already knows the recipient VASP address and an identifier for the recipient at their VASP. The sender VASP contacts the recipient VASP and shares the sender's identifier, a reference ID, and the recipient's identifier. 
+Upon conclusion, the sender VASP transmits to the blockchain a transaction with the reference ID, completing the protocol without revealing any potentially identifiable user information publicly on-chain.
+
+Reference IDs are unique between a pair of VASPs, `VASP A` and `VASP B`,  so if a reference ID `r` was used in a previous transaction between the two VASPs, regardless of which one initiated it, `r` cannot be reused for any future transactions between `VASP A` and `VASP B`. 
+However, the same `r` value can be used between a different VASP pair, `VASP A` and `VASP C`. VASPs may choose to keep reference IDs globally unique across all VASPs as well.
```
sunmilee, line 23: Also added more details on ref ID uniqueness @aching

davidiw (commented):
forgot to submit this earlier.

dips/dip-183.md
davidiw: Personally I think the first sentence was good enough, but I'd rather over specify than have confusion :P

davidiw (approved):

PR #190, opened 2021-06-18
[DIP 169] remove section on designated dealers

emmazzz:
This PR removes references to designated dealers in DIP 169 since the use case is no longer applicable.

### DIFF
--- a/dips/dip-169.md
+++ b/dips/dip-169.md
@@ -25,27 +25,6 @@ Currently in the Diem Framework, a transaction acts on behalf of a single on-cha
 
 # Motivation/Use Cases
 
-## Minting Directly to VASPs
-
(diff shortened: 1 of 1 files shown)


dahliamalkhi (commented):
I wonder: Instead of removing the example entirely from the DIP, could we reword it a bit? we can say :

coins can minted only into an account that has minting capability and controlled by T&C (currently called a DD but may change in the future). However, the minting account will frequently provide mint/sell services to VASPs and want to transfer coins to the VASP, rather than keep them on the balance even for a short duration. Hence, multi-agent allows to bundle the two actions into one. 

If people prefer to erase the example completely, I'll go ahead and approve the PR.

aching:
I am comfortable removing that section unless someone really wants it.  There are several other good examples in DIP-169.

dahliamalkhi (approved):

PR #193, opened 2021-06-25
DIP10 command name fix

sunmilee:
Changing name to match DIP183

### DIFF
--- a/dips/dip-10.md
+++ b/dips/dip-10.md
@@ -72,7 +72,7 @@ The format of the command is:
    "_ObjectType": "CommandRequestObject",
     "command_type": "SenderInitiatedReferenceIDExchange",
     "command": {
-	    "_ObjectType": "ReferenceIDCommand",
+	    "_ObjectType": "SenderInitiatedReferenceIDExchange",
 	    "sender": "alice@avasp",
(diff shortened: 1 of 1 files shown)


dips/dip-10.md
```diff
@@ -72,7 +72,7 @@ The format of the command is:
    "_ObjectType": "CommandRequestObject",
     "command_type": "SenderInitiatedReferenceIDExchange",
     "command": {
-	    "_ObjectType": "ReferenceIDCommand",
+	    "_ObjectType": "SenderInitiatedReferenceIDExchange",
```
xli, line 75: I thought you proposed to use `SenderInitiatedDiemIDExchange`

davidiw: DiemID is one of many identifiers that could be used, so it isn't quite accurate.

davidiw (approved):

dips/dip-10.md
sunmilee: I heard that we want to move away from using Diem ID everywhere :D I don't mind either `SenderInitiatedReferenceIDExchange` or `SenderInitiatedDiemIDExchange`, it's more general vs more specific.. @davidiw can help make final call please

PR #198, opened 2021-07-07
DIP183 Last Call

sunmilee:


davidiw:
Considering the general direction here, I'd actually propose a bit of rework in DIP-183 in particular with examples and test data.

Examples which could go into the appendix include various different strategies for exchanging a reference id:
* Subaddresses
* PayString
* Other naming services
* Leveraging Diem domain identifiers in addition to other services

Then we could avoid introducing a new naming construct that is Diem specific and altogether unnecessary.

In addition, we should have good test cases for verifying correctness of the exchange.


For DIP-10, it doesn't seem to focus on the intended point which is just the Diem domain.

Issue #204, opened 2021-08-11
State Sync V2


JoshLind:
This issue is to track the development of the DIP for state sync v2. A pull request with a draft DIP will follow shortly. The TL;DR for those already interested:

* *Problem*: State sync v1.1 [1] only offers a single primitive strategy for keeping nodes up-to-date. This strategy is _slow_ (i.e., it requires syncing and executing all transactions from genesis), _expensive_ (i.e., it requires downloading, executing and storing the entire transaction history), _inflexible_ (i.e., it is unable to adapt to different use-cases and requirements) and _brittle_ (i.e., the implementation tightly couples Diem components and Diem nodes).
* *Solution*: We propose a new state sync framework (_state sync v2_) that supports multiple state sync strategies based on use-cases. The strategies trade-off security, performance and resources. Strategies include: syncing from genesis, syncing directly from a waypoint and syncing without executing. The new framework decouples intra-node components (e.g., state sync and storage) and removes coupling between Diem nodes (e.g., allowing additional strategies to be introduced without breaking backward compatibility).

[1] https://github.com/diem/diem/issues/6795

PR #205, opened 2021-08-11
[Dip: 205] Add an initial draft DIP for State Sync v2.

JoshLind:
This PR is for an initial draft DIP of State Sync v2 (see issue: https://github.com/diem/dip/issues/204).

### DIFF
--- a/dips/dip-205.md
+++ b/dips/dip-205.md
@@ -1,0 +1,163 @@
+---
+dip: 205
+title: State Sync v2: Strategies & Framework
+authors: Joshua Lind (@joshlind)
+status: Draft
+type: Standard
+created: 08/11/2021
+---
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


JoshLind:
CC' @davidiw, @aching

Note: I can't add reviewers to this PR -- I guess this is intentional?

davidiw:
I wonder these things too. I think there are limited permissions. I'll poke at it later.

gregnazario:
@davidiw @aching Who's the DIP manager for this DIP?  And do we need this DIP?

davidiw:
Informational DIPs are welcome. I do think we need a storage service DIP and if we aren't going to do storage service in the first pass, we should make a state sync v2 server DIP.

davidiw:
I can take a look at this

PR #208, opened 2021-09-02
[DIP-201] amend DiemBFT v4 pdf

zekun000:


aching (approved):
Thank you!

aching:
@zekun000  Actually, does the link work for you?  It failed for me.  I wonder if it's because of "DiemBFT-V4.pdf" instead of "DiemBFT-v4.pdf"?

aching:
After the casing fix, the link works!  Thanks!

PR #210, opened 2021-09-14
[DIP-201] remove testing section and reword

zekun000:


dips/dip-201.md
```diff
@@ -65,14 +65,6 @@ Commit Rule:
 Switching from existing DiemBFT to the newer version requires reconfiguration. An on-chain consensus config will be added to support the transition from 3-chain to 2-chain at an epoch boundary.
```
dahliamalkhi, line 65: could you explain:
1. In TwoChainTimeoutCertificate, you say only one signature is kept (as optimization. What is the BtreeMap used for then? is it expected that all nodes sending a timeout sign the same TwoChainTimeout struct? 
2. "the signature is signed on the derived struct" -- which "the signature"? also, add period at end of sentence.
3. can you explain where two-chain timeouts are used in the voting/commit rules?

dahliamalkhi (changes requested):

dips/dip-201.md
zekun000: 1. 
> What is the BtreeMap used for then?

Only one *QuorumCert` is kept instead of 2f+1, not only one signature. 
BTreeMap keeps 2f+1 signatures and the highest qc round from the node. 

> is it expected that all nodes sending a timeout sign the same TwoChainTimeout struct?

No, they sign on the same epoch and round but with its own highest qc. (In TimeoutCert, we only keep the highest qc among the 2f+1 timeout).

2. "the signature" on for TwoChainTimeout.

3. only TwoChainTimeoutCertificate is used in new rules, it's used as `tc` in the rules e.g. (block.round == tc.round + 1 && block.qc.round >= tc.highest_hqc_round)

zekun000: updated some words

dips/dip-201.md
```diff
@@ -39,8 +39,11 @@ pub struct TwoChainTimeoutCertificate {
 }
 ```
 
-TwoChainTimeoutCertificate only carries a single quorum cert instead of 2f+1 as an optimization to save both bandwidth and verification cost.
-It keeps the TwoChainTimeout with the highest quorum cert and the signature is signed on the derived struct
+The signature for TwoChainTimeout is signed on the derived struct TimeoutSigningRepr which uses the highest qc round instead of the actual QuorumCert
```
dahliamalkhi, line 42:
```suggestion
The signature for TwoChainTimeout is done **indirectly**, on the derived struct TimeoutSigningRepr which uses the highest qc round instead of an actual QuorumCert 
```

dahliamalkhi (changes requested):

sasha8 (approved):

dahliamalkhi (approved):

Issue #213, opened 2021-11-24
decoupled execution from consensus


zekun000:
placeholder for now

dahliamalkhi:
I will be DIP Manager for this topic.

PR #214, opened 2021-11-24
[dip-213] decoupled execution

zekun000:
for #213

### DIFF
--- a/dips/dip-213.md
+++ b/dips/dip-213.md
@@ -1,0 +1,96 @@
+---
+dip: 213
+title: Decoupled execution 
+authors: Zekun Li (@zekun000), Yu Xia (@yuxiamit)
+status: Draft
+type: Standard
+created: 08/09/2021
+last updated: 09/09/2021
+issue: https://github.com/diem/dip/issues/213
+---
+
+# Summary
+
+The current [DiemBFTv4](https://developers.diem.com/papers/diem-consensus-state-machine-replication-in-the-diem-blockchain/2021-08-17.pdf) consensus core engine forms agreement on transaction ordering jointly with execution results.
+That is, a transaction (block) is **finalized** in the history of transactions at some position 'x' at the same time that validators vote on a ledger state `LedgerInfo` resulting from executing the history-prefix ending with the transaction (block) at position 'x'. The upside of this approach is simplicity.
+The down-side is that the coupling execution with ordering stalls each other and limits throughput.
+This DIP proposes to separate them to unlock better throughput without compromising security.
+
+```
+Current paradigm: Order → Execute → Commit → Order → Execute → Commit
+
+Proposed paradigm: Order → Order → Order
+                     ↓        ↓
+                     Execute → Execute
+                        ↓        ↓
+                       Commit → Commit
+```
+
+If the time of ordering, execution and commit for a block is O, E, C correspondingly, the throughput should change from 1/(O + E + C) to 1/max(O, E, C) which not only unlocks throughput benefits immediately,
+but makes bottlenecks more visible and could guide future optimizations.
+
+# Description
+
+In [DiemBFTv4](https://developers.diem.com/papers/diem-consensus-state-machine-replication-in-the-diem-blockchain/2021-08-17.pdf), a block (proposal) goes through different steps before it's finalized. The steps are 
+```
+Consensus stage
+    1. Proposed
+    2. Executed
+    3. Voted (including execution result)
+    4. QuorumCertified
+    5. 2-chain Certified (Finalized)
+```
+
+The proposed change is to decouple execution from consensus and pipeline the process with different stages. A block would go through
+4 stages each with its own steps:
+```
+1. Consensus stage 
+    * Proposed
+    * Voted (no execution result)
+    * QuorumCertified
+    * 2-chain Certified (Ordered)
+2. Execution stage
+3. Voting stage (for execution result)
+4. Commit stage (Finalized)
+```
+The stages run in parallel, utilizing available compute resources to increase overall throughput. 
+That is, the system could commit B1, sign B2, execute B3, and order B4, all at the same time. Note, while throughput increases due to parallel execution, latency of an individual transaction (block) finalization does not decrease, as the stages are inherently sequential. In fact, individual transaction latency may slightly increase since commit occurs less "eagerly".
+
+# Required Changes
+
+## Consensus
+`StateComputer` is the trait that consensus uses for `compute`, `commit` or `sync_to` blocks.
+An `ordering_state_computer` is implemented to bypass execution and send blocks to next stage when ordered.
+
+## BlockStateManager
+A `BlockStateManager` is implemented to manage different stages of ordered blocks, it's a queue of blocks with stage markers.
+- `Ordered` block is sent to execution and advanced to `Executed` after receiving execution result.
+- `Executed` block is sent to safety rules and advanced to `Signed` after receiving a signature.
+- Signature on `Signed` block is broadcasted to all validators. At all stages except `Aggregated`, signatures from other validators on the block are collected.
+- A block in either `Executed` or `Signed` stage is advanced to `Aggregated` after receiving a quorum of signatures.
+- `Aggregated` block is popped from the queue and sent to persistent storage.
+ 
+## Sync
+When the node receives the SyncInfo carried in Proposal/Vote messages, it checks the highest round numbers compared to local round and may decide to fast-forward via state sync protocol. It's triggered if the gap between
+the local highest round of ordered block and the remote highest round of a committed block becomes large (current threshold is 10). Upon state sync, 
+the node fast-forwards to the same state (both ledger state and pending blocks) as the remote peer. 
+
+## Backpressure
+To prevent consensus-ordering going too fast and creating an ever-growing backlog for other stages, back pressure is implemented to stop ordering from making progress if the gap between the highest committed block and highest ordered block is large.
+
+## Reconfiguration
+Recall that a reconfiguration transaction must be the last transaction of an epoch. Before this proposal, consensus-ordering is aware of 
+reconfiguration and leader proposes empty suffix. After the proposal, execution on each node recognizes this type of transaction and BlockStateManager stops processing any blocks succeeding the reconfiguration block.
+Ordering stage contineus moving forward until backpressure is triggered. Ordered blocks after the reconfiguration block are discarded and transactions inside those blocks would be retried in next epoch.
+Imagine a few block batches certified by ledger info, [1, 2] [r, 3], [4], [5, 6], r represents a block that contains reconfiguration transaction. The ledger info is aggregated on block 3 (transactions in 3 are skipped) and 4, 5, 6 are discarded.
+After the reconfiguration block gets committed, the old epoch instance is dropped, a new epoch is instantiated via a virtual genesis block derived from the last ledger info in previous epoch.
+
+## Upgrade
+To enable the feature in an existing blockchain requires coordination via reconfiguration, on-chain consensus config would be updated to a newer version supporting the upgrade protocol.
+
+## Client
+The finality proof is aggregated in the same format as before (LedgerInfoWithSignatures), so this change is client agnostic.
+
+# Future opportunities
+The current suggested implementation implements execution signature aggregation via a simple broadcast, a leader based mechanism could be implemented to reduce the network cost
+by compromising one-hop latency.


dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+issue: https://github.com/diem/dip/issues/213
+---
+
+# Summary
+
+The current consensus agrees on transaction ordering together with execution results. The good part is the simplicity that enables us to implement and harden within short time period. 
+The bad part is that the coupling between execution and ordering stall each other and limit the performance. 
+This DIP proposes to separate them to unlock better throughput without compromising security.
+
+If the time of ordering, execution and commit for a block is O, E, C correspondingly, the throughput should change from 1/(O + E + C) to 1/min(O, E, C) which not only unlocks wins immediately
```
dahliamalkhi: can you add the nice visual you had in internal design docs ?

Current paradigm: Order → Execute → Commit → Order → Execute → Commit

Decoupled paradigm: Order → Order → Order
↓ ↓
Execute → Execute
↓ ↓
Commit → Commit

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+status: Draft
+type: Standard
+created: 08/09/2021
+last updated: 09/09/2021
+issue: https://github.com/diem/dip/issues/213
+---
+
+# Summary
+
+The current consensus agrees on transaction ordering together with execution results. The good part is the simplicity that enables us to implement and harden within short time period.
```
dahliamalkhi:
```suggestion
The current [DiemBFTv4](url) consensus core engine forms agreement on transaction ordering jointly with execution results. That is, a transaction (block) is **finalized** in the history of transactions at some position 'x' at the same time that validators vote on a ledger state `LedgerInfo` resulting from executing the history-prefix ending with the transaction (block) at position 'x'. The upside of this approach is simplicity. 
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+type: Standard
+created: 08/09/2021
+last updated: 09/09/2021
+issue: https://github.com/diem/dip/issues/213
+---
+
+# Summary
+
+The current consensus agrees on transaction ordering together with execution results. The good part is the simplicity that enables us to implement and harden within short time period. 
+The bad part is that the coupling between execution and ordering stall each other and limit the performance.
```
dahliamalkhi:
```suggestion
The down-side is that the coupling execution with ordering stalls each other and limits throughput. 
```

dips/dip-213.md
dahliamalkhi:
```suggestion
If the time of ordering, execution and commit for a block is O, E, C correspondingly, the throughput should change from 1/(O + E + C) to 1/min(O, E, C) which not only unlocks throughput benefits immediately, 
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+# Summary
+
+The current consensus agrees on transaction ordering together with execution results. The good part is the simplicity that enables us to implement and harden within short time period. 
+The bad part is that the coupling between execution and ordering stall each other and limit the performance. 
+This DIP proposes to separate them to unlock better throughput without compromising security.
+
+If the time of ordering, execution and commit for a block is O, E, C correspondingly, the throughput should change from 1/(O + E + C) to 1/min(O, E, C) which not only unlocks wins immediately 
+but make future bottlenecks more visible and could guide optimizations.
```
dahliamalkhi:
```suggestion
but makes bottlenecks more visible and could guide future optimizations.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Description
+
+A block (proposal) goes through different steps before it's finalized. In DiemBFTv4, the steps are
```
dahliamalkhi: add ref to 'state machine replication' whitepaper

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+# Description
+
+A block (proposal) goes through different steps before it's finalized. In DiemBFTv4, the steps are 
+```
+Consensus stage
+    1. Proposed
+    2. Executed
+    3. Voted (including execution result)
+    4. QuorumCertified
+    5. 2-chain Certified (Committed)
```
dahliamalkhi: let's use finalized/committed consistently throughout the DIP. Above, "finalized" is used (and I also used in my edits), here "committed".

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+1. Consensus stage 
+    * Proposed
+    * Voted (no execution result)
+    * QuorumCertified
+    * 2-chain Certified (Ordered)
+2. Execution stage
+3. Voting stage (for execution result)
+4. Commit stage
+```
+The stages run in parallel to achieve the best resource utilization, the system could commit B1, sign B2, execute B3, and order B4 in the same time.
```
dahliamalkhi:
```suggestion
The stages run in parallel, utilizing available compute resources to increase overall throughput. That is, the system could commit B1, sign B2, execute B3, and order B4, all at the same time. Note, while throughput increases due to parallel execution, latency of an individual transaction (block) finalization does not decrease, as the stages are inherently sequential. In fact, individual transaction latency may slightly increase since commit occurs less "eagerly". 
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Required Changes
+
+## Consensus
+`StateComputer` is the trait that consensus uses for `compute`, `commit` or `sync_to` blocks.
+A simple `ordering_state_computer` is implemented to bypass execution and send blocks to next stage when ordered.
```
dahliamalkhi:
```suggestion
An `ordering_state_computer` is implemented to bypass execution and send blocks to next stage when ordered.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+
+## BufferManager
+A `BufferManager` is implemented to manage different stages of ordered blocks, it's a queue of blocks with stages markers.
```
dahliamalkhi:
```suggestion
A `BufferManager` is implemented to manage different stages of ordered blocks, it's a queue of blocks with stage markers.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+```
+The stages run in parallel to achieve the best resource utilization, the system could commit B1, sign B2, execute B3, and order B4 in the same time.
+
+# Required Changes
+
+## Consensus
+`StateComputer` is the trait that consensus uses for `compute`, `commit` or `sync_to` blocks.
+A simple `ordering_state_computer` is implemented to bypass execution and send blocks to next stage when ordered.
+
+## BufferManager
```
dahliamalkhi: I don't like this name, this seems to have very little to do with buffering?

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+A `BufferManager` is implemented to manage different stages of ordered blocks, it's a queue of blocks with stages markers.
+- `Ordered` block is sent to execution and advanced to `Executed` after receiving execution result.
+- `Executed` block is sent to safety rules and advanced to `Signed` after receiving signature.
```
dahliamalkhi:
```suggestion
- `Executed` block is sent to safety rules and advanced to `Signed` after receiving a signature.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+- Signature on `Signed` block is broadcasted to every validators until it's advanced to `Aggregated`.
```
dahliamalkhi:
```suggestion
- Signature on `Signed` block is broadcasted to all validators. Both at this stage and the Executed stage, signatures from other validators on the block are collected.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+- `Executed` and `Signed` block is advanced to `Aggregated` after receiving enough signatures.
```
dahliamalkhi:
```suggestion
-  A block in either `Executed` or `Signed` stage is advanced to `Aggregated` after receiving a quorum of signatures.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+- `Aggregated` block is popped from the queue and sent to persistent storage.
+ 
+## Sync
+When the node is far from current proposed block, it may decide to fast-forward via state sync protocol. It's triggered by the difference between
```
dahliamalkhi: Not clear. please explain.

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+local ordered round (the highest round of ordered blocks) and remote committed round (the highest round of committed block). Upon state sync, 
+the node needs to fetch blocks chaining from committed round to ordered round + 2 (which carries the ordering certificate) instead of ordered round to ordered round + 2 before.
+Block retrieval is improved to support chunked requests.
+
+## Backpressure
+To prevent consensus going too fast and create ever-growing backlog for other stages, back pressure is implemented to stop consensus making progress if the difference between committed block and ordered block is large.
```
dahliamalkhi:
```suggestion
To prevent consensus-ordering going too fast and create ever-growing backlog for other stages, back pressure is implemented to stop ordering from making progress if the gap between the highest committed block and highest ordered block is large.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Reconfiguration
+A reconfiguration transaction is the last transaction of an epoch. Execution recognizes this type of transaction and buffer manager would stop processing any blocks after the reconfiguration block.
+Consensus would stop once the backpressure is triggered, blocks after the reconfiguration block would be discarded and transactions would be retried in next epoch.
+After the reconfiguration block gets committed, new epoch would be instantiated and old epoch instance would bd dropped.
```
dahliamalkhi:
```suggestion
After the reconfiguration block gets committed, a new epoch is instantiated the old epoch is dropped.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+When the node is far from current proposed block, it may decide to fast-forward via state sync protocol. It's triggered by the difference between
+local ordered round (the highest round of ordered blocks) and remote committed round (the highest round of committed block). Upon state sync, 
+the node needs to fetch blocks chaining from committed round to ordered round + 2 (which carries the ordering certificate) instead of ordered round to ordered round + 2 before.
+Block retrieval is improved to support chunked requests.
+
+## Backpressure
+To prevent consensus going too fast and create ever-growing backlog for other stages, back pressure is implemented to stop consensus making progress if the difference between committed block and ordered block is large.
+
+## Reconfiguration
+A reconfiguration transaction is the last transaction of an epoch. Execution recognizes this type of transaction and buffer manager would stop processing any blocks after the reconfiguration block.
```
dahliamalkhi:
```suggestion
Recall that a reconfiguration transaction must be the last transaction of an epoch. Execution recognizes this type of transaction and BufferManager stops processing blocks succeeding the reconfiguration block.
```

dahliamalkhi: what about the empty/special block that must follow a reconfiguration block to reach a 2-chain commit?

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+Consensus would stop once the backpressure is triggered, blocks after the reconfiguration block would be discarded and transactions would be retried in next epoch.
```
dahliamalkhi: not clear. Please explain

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+After the reconfiguration block gets committed, new epoch would be instantiated and old epoch instance would bd dropped.
+
+## Upgrade
+Switching the protocol requires reconfiguration, on-chain consensus config is updated to newer version to support the upgrade.
```
dahliamalkhi:
```suggestion
Switching to the protocol requires reconfiguration: on-chain consensus config would be updated to a newer version supporting the updated protocol.
```

dips/dip-213.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Client
+The finality proof is aggregated in the same format as before (LedgerInfoWithSignatures), so this change is client agnostic.
+
+# Future opportunities
+The execution signature aggregation is implemented as simple broadcast, a leader based mechanism could be implemented to reduce the network cost
```
dahliamalkhi:
```suggestion
The current suggested implementation implements execution signature aggregation via a simple broadcast. In the future, a leader based mechanism could be implemented to reduce the network cost
```

dahliamalkhi (changes requested):
wonderful proposal! see stylistic/clarification comments inline.

dips/dip-213.md
zekun000: I don't like it either, it's just what we use now, happy to rename it to something better.

zekun000: I was proposing StateManager but State seems also an over-used term, would love suggestions!

dips/dip-213.md
zekun000: those empty blocks won't be committed (both before and after), they're just used to facilitate the commit of reconfiguration block.

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+Current paradigm: Order → Execute → Commit → Order → Execute → Commit
+
+Proposed paradigm: Order → Order → Order
+                     ↓        ↓
+                     Execute → Execute
+                        ↓        ↓
+                       Commit → Commit
+```
+
+If the time of ordering, execution and commit for a block is O, E, C correspondingly, the throughput should change from 1/(O + E + C) to 1/min(O, E, C) which not only unlocks throughput benefits immediately,
```
dahliamalkhi: shouldn't tput be 1/max(O, E, C)?

dips/dip-213.md
dahliamalkhi: it's really a block state-machine, but that's of course overloaded.

how about ConsensusStageManager or BFTStageManager or SMRStageManager?

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+## BufferManager
+A `BufferManager` is implemented to manage different stages of ordered blocks, it's a queue of blocks with stage markers.
+- `Ordered` block is sent to execution and advanced to `Executed` after receiving execution result.
+- `Executed` block is sent to safety rules and advanced to `Signed` after receiving a signature.
+- Signature on `Signed` block is broadcasted to all validators. At all stages except `Aggregated`, signatures from other validators on the block are collected.
+- A block in either `Executed` or `Signed` stage is advanced to `Aggregated` after receiving a quorum of signatures.
+- `Aggregated` block is popped from the queue and sent to persistent storage.
+ 
+## Sync
+When the node is far from current proposed block, it may decide to fast-forward via state sync protocol. It's triggered if the gap between
```
dahliamalkhi:
```suggestion
When a node's local blockchain tail is far from current proposed block, it may decide to fast-forward via state sync protocol. It's triggered if the gap between
```

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+local highest round of ordered block and remote highest round of committed block becomes large (current threshold is 10). Upon state sync,
```
dahliamalkhi:
```suggestion
the local highest round of ordered block and the highest round of a remote committed block becomes large (current threshold is 10). Upon state sync, 
```

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+- `Ordered` block is sent to execution and advanced to `Executed` after receiving execution result.
+- `Executed` block is sent to safety rules and advanced to `Signed` after receiving a signature.
+- Signature on `Signed` block is broadcasted to all validators. At all stages except `Aggregated`, signatures from other validators on the block are collected.
+- A block in either `Executed` or `Signed` stage is advanced to `Aggregated` after receiving a quorum of signatures.
+- `Aggregated` block is popped from the queue and sent to persistent storage.
+ 
+## Sync
+When the node is far from current proposed block, it may decide to fast-forward via state sync protocol. It's triggered if the gap between
+local highest round of ordered block and remote highest round of committed block becomes large (current threshold is 10). Upon state sync, 
+the node fast-forward to the same state (both ledger state and pending blocks) as the remote peer.
```
dahliamalkhi:
```suggestion
the node fast-forwards to the same state (both ledger state and pending blocks) as the remote peer. 
```

dips/dip-213.md
dahliamalkhi: what do you mean by "the node is far"? do you mean it local blockchain tail?

and how does the node learn about the "gap" without being in state-sync already? this seems circular, first you need to state-sync in order to discover the gap and then initiate a state-sync ..

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+- A block in either `Executed` or `Signed` stage is advanced to `Aggregated` after receiving a quorum of signatures.
+- `Aggregated` block is popped from the queue and sent to persistent storage.
+ 
+## Sync
+When the node is far from current proposed block, it may decide to fast-forward via state sync protocol. It's triggered if the gap between
+local highest round of ordered block and remote highest round of committed block becomes large (current threshold is 10). Upon state sync, 
+the node fast-forward to the same state (both ledger state and pending blocks) as the remote peer. 
+
+## Backpressure
+To prevent consensus-ordering going too fast and create ever-growing backlog for other stages, back pressure is implemented to stop ordering from making progress if the gap between the highest committed block and highest ordered block is large.
```
dahliamalkhi:
```suggestion
To prevent consensus-ordering going too fast and creating an ever-growing backlog for other stages, back pressure is implemented to stop ordering from making progress if the gap between the highest committed block and highest ordered block is large.
```

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Reconfiguration
```
dahliamalkhi, line 81: please explain: why are we describing this here, and what is the connection to decoupled execution?

dips/dip-213.md
```diff
@@ -0,0 +1,94 @@
...
+Recall that a reconfiguration transaction must be the last transaction of an epoch. Execution recognizes this type of transaction and BufferManager stops processing any blocks succeeding the reconfiguration block.
+Ordered blocks after the reconfiguration block are discarded and transactions inside those blocks would be retried in next epoch.
+After the reconfiguration block gets committed, a new epoch is instantiated, the old epoch instance is dropped.
+
+## Upgrade
```
dahliamalkhi, line 88: same comment as above.

dahliamalkhi (changes requested):

dips/dip-213.md
zekun000: ah, good catch!

dips/dip-213.md
zekun000: the node learns about the gap by checking round numbers from SyncInfo (all highest certs) (which is carried in proposal/vote/timeout etc) and decides whether to initiate state-sync

dips/dip-213.md
zekun000: before decoupled-execution, proposer is aware of reconfiguration and propose empty suffix (also enforced by voters).
after the change, each node decides to skip the suffix during execution (ordering is not aware of reconfiguration anymore).

dips/dip-213.md
zekun000: this is about the typical upgrade plan to enable the feature?

dips/dip-213.md
dahliamalkhi: First, it's definitely worth explaining this carefully here.

Second, this concerns me a bit: Who and when creates a new epoch genesis then?

dips/dip-213.md
dahliamalkhi: got it. please incorporate into the DIP?

dips/dip-213.md
dahliamalkhi: got it. please incorporate the explanation here.

dips/dip-213.md
zekun000: the (virtual) genesis block is derived from the ledger info from the previous epoch. the deriving function remains unchanged.
with block batches [1 -> 2] -> [r -> 3] -> [4] -> [5, 6], the ledger info would be aggregated on 3 (txns in 3 are skipped).

dahliamalkhi: Got it, but all worth elaborating here -- 4, 5, 6 -- skipped? when does a leader propose in the new epoch? etc etc

zekun000: yes, 4, 5, 6 are all skipped, nodes block on aggregating ledger info on 3 (periodically broadcast signature until commit).  I was trying to only discuss the delta here though - how is genesis derived, when does leader enter new epoch, how do nodes catch up etc remains unchanged

dahliamalkhi: I don't think it's the same nor trivial. 

IIUC, a leader waits for commit votes on 3 and then discards all blocks after 3, forms an epoch genesis, and proposes it, right? Let's say a few more words to this effect ;-)

dips/dip-213.md
```diff
@@ -0,0 +1,96 @@
...
+the node fast-forwards to the same state (both ledger state and pending blocks) as the remote peer. 
+
+## Backpressure
+To prevent consensus-ordering going too fast and creating an ever-growing backlog for other stages, back pressure is implemented to stop ordering from making progress if the gap between the highest committed block and highest ordered block is large.
+
+## Reconfiguration
+Recall that a reconfiguration transaction must be the last transaction of an epoch. Before this proposal, consensus-ordering is aware of 
+reconfiguration and leader proposes empty suffix. After the proposal, execution on each node recognizes this type of transaction and BlockStateManager stops processing any blocks succeeding the reconfiguration block.
+Ordered blocks after the reconfiguration block are discarded and transactions inside those blocks would be retried in next epoch.
+Imagine a few block batches [1, 2] [r, 3], [4], [5, 6], r represents a block has reconfiguration transaction. The ledger info is aggregated on block 3 (transactions in 3 are skipped) and 4, 5, 6 are discarded.
```
dahliamalkhi: what is the purpose of the bracket notation? what do you mean by "block batches" (transactions batched into blocks?)

dahliamalkhi:
```suggestion
Imagine a few block batches [1, 2] [r, 3], [4], [5, 6], r represents a block that contains a reconfiguration transaction. The ledger info is aggregated on block 3 (transactions in 3 are skipped) and 4, 5, 6 are discarded.
```

dips/dip-213.md
```diff
@@ -0,0 +1,96 @@
...
+After the reconfiguration block gets committed, the old epoch instance is dropped, a new epoch is instantiated via a virtual genesis block derived from the last ledger info in previous epoch.
+
+## Upgrade
+To enable the feature in existing blockchain requires coordination via reconfiguration, on-chain consensus config would be updated to a newer version supporting the upgrade protocol.
```
dahliamalkhi:
```suggestion
To enable the feature in an existing blockchain requires coordination via reconfiguration, on-chain consensus config would be updated to a newer version supporting the upgrade protocol.
```

dips/dip-213.md
```diff
@@ -0,0 +1,96 @@
...
+When the node receives the SyncInfo carried in Proposal/Vote messages, it checks the highest round numbers compared to local round and may decide to fast-forward via state sync protocol. It's triggered if the gap between
+the local highest round of ordered block and the remote highest round of a committed block becomes large (current threshold is 10). Upon state sync, 
+the node fast-forwards to the same state (both ledger state and pending blocks) as the remote peer. 
+
+## Backpressure
+To prevent consensus-ordering going too fast and creating an ever-growing backlog for other stages, back pressure is implemented to stop ordering from making progress if the gap between the highest committed block and highest ordered block is large.
+
+## Reconfiguration
+Recall that a reconfiguration transaction must be the last transaction of an epoch. Before this proposal, consensus-ordering is aware of 
+reconfiguration and leader proposes empty suffix. After the proposal, execution on each node recognizes this type of transaction and BlockStateManager stops processing any blocks succeeding the reconfiguration block.
```
dahliamalkhi, line 83: does this mean that the consensus stage continues moving forward, oblivious of the reconfiguration transaction, until either execution catches up with the reconfig tx or back-pressure prevents progress? if yes, perhaps add this clarification to the text?

dahliamalkhi (changes requested):

dips/dip-213.md
zekun000: it's a batches of blocks that certified by a ledger info, not every block gets a ledger info in case of timeout

dips/dip-213.md
zekun000: it moves forward until it hits backpressure and then blocks until next epoch

dips/dip-213.md
```diff
@@ -0,0 +1,96 @@
...
+## Consensus
+`StateComputer` is the trait that consensus uses for `compute`, `commit` or `sync_to` blocks.
+An `ordering_state_computer` is implemented to bypass execution and send blocks to next stage when ordered.
+
+## BlockStateManager
+A `BlockStateManager` is implemented to manage different stages of ordered blocks, it's a queue of blocks with stage markers.
+- `Ordered` block is sent to execution and advanced to `Executed` after receiving execution result.
+- `Executed` block is sent to safety rules and advanced to `Signed` after receiving a signature.
+- Signature on `Signed` block is broadcasted to all validators. At all stages except `Aggregated`, signatures from other validators on the block are collected.
+- A block in either `Executed` or `Signed` stage is advanced to `Aggregated` after receiving a quorum of signatures.
```
mimoo, line 70: I would suggest "certification of correct execution", as 2f+1 have agreed that this is the correct execution output

mimoo (commented):
wow this is great, is this implemented already? Are there any downsides of this approach?

zekun000:
@mimoo yes it's already implemented. one trade-off is that it adds one hop latency due to the extra execution certification

mimoo:
oh right, makes sense! thanks

